PR #5, opened 2023-07-07
BA-1085: Initiate baseapp-auth

Janekk:
**Acceptance Criteria** 

- This package should be a subfolder in https://github.com/silverlogic/baseapp-backend
- move all of our authentication system into this page
- move all git history of moved files as well
- including the user model
- provide abstract user model
- needs to allow to swap user model by the project
- use JWT for token instead
- store full private user serializer, to avoid FE hitting BE to get info about logged in user
- make sure frontend is compatible
- review change password flow


baseapp-auth/baseapp_auth/baseapp_referrals
nossila, line 1: this symlink needs to be relative I think

baseapp-auth/testproject/tests/fixtures.py
nossila, line 1: Most of this fixtures are on baseapp_core already, you could import it from there instead, so we don't have duplicated code

baseapp-auth/testproject/tests/helpers.py
nossila, line 1: same as this helpers, its on baseapp_core

baseapp-auth/testproject/tests/integration/test_migration_health_check.py
nossila, line 1: this is also part of baseapp_core

baseapp-auth/testproject/tests/mixins.py
nossila, line 1: this as well, part of baseapp_core

baseapp-auth/testproject/tests/unit/test_celery_beat_health_check.py
nossila, line 1: this as well, part of baseapp_core

baseapp-core/baseapp_core/settings/env.py
nossila, line 1: this also belongs in baseapp_core in my opinion, but its not there yet

nossila (changes requested):

baseapp-core/baseapp_core/tests/fixtures.py
```diff
@@ -9,9 +9,9 @@
 from asgiref.sync import sync_to_async
 from rest_framework.test import APIClient
 
-from ..graphql.testing import graphql_query
+import testproject.tests.factories as f
```
nossila, line 12: I think this can break for projects importing fixtures from this file, I think it shouldn't import factories from the testproject here

baseapp-auth/manage.py
nossila, line 1: I think this should be moved into the testproject

baseapp-auth/baseapp_auth/rest_framework/register/serializers.py
```diff
@@ -0,0 +1,39 @@
+from django.contrib.auth import get_user_model
+from django.utils.translation import gettext as _
+
+from baseapp_referrals.models import UserReferral
```
pablogadhi, line 4: Something for the future, but I think we should decouple this module from any logic on the referrals module.

pablogadhi (approved):
Tested the JWT flow and decoded the token to see the data. It's looking awesome!
image

baseapp-auth/baseapp_auth/baseapp_referrals
Janekk: resolved in https://github.com/silverlogic/baseapp-backend/pull/5/commits/603909e8d691bcc96c5ad30dd9ffabf6f1c403c1

nossila (approved):

PR #12, opened 2023-07-18
BA-1084-add-release

do-tsl:


baseapp-core/setup.py
```diff
@@ -1,3 +1,7 @@
 from setuptools import setup
 
-setup()
+setup(
+    name="baseapp-core",
+    packages=["baseapp-core"],
+    version="0.1",
```
Janekk, line 6: should we keep version from `setup.cfg`? 

```
version = 0.1.5
```

Janekk (changes requested):

pablogadhi (approved):

baseapp-core/setup.py
do-tsl: I guess we do that once we start using the real account, right now we are on test one

PR #21, opened 2023-08-07
BA-1128: Deployment of pypi packages to production

pablogadhi:


deboracosilveira (approved):

Janekk (changes requested):
could we keep test-publish workflow in its own file?

Janekk (approved):

PR #56, opened 2023-12-29
Sms templates

Ronan-Fernandes:
Create Sms Template model and its utils to be able to store sms text messages in a simpler way


Janekk:
Could we please update the readme of the `baseapp-email-templates` project?

PR #60, opened 2024-01-09
Rename Email Templates

Ronan-Fernandes:
Rename Email templates to Message Templates


pt-tsl (approved):

.github/workflows/project-workflow-old-renamed-packages.yml
nossila, line 1: why is this file necessary? seems like we can just remove `baseapp-email-templates` from both `github-actions.yml` and `publish-actions.yml`

baseapp-email-templates/setup.py
```diff
@@ -1,12 +1,18 @@
...
+
 setup(
     name="baseapp-email-templates",
-    packages=find_packages(),
-    version="0.1",
+    description="baseapp-email-templates is now baseapp-message-templates",
     long_description=long_description,
     long_description_content_type="text/markdown",
+    version=VERSION,
+    install_requires=["baseapp-message-templates"],
```
nossila, line 16: the package don't need to require itself, we can remove `install_requires`

baseapp-message-templates/README.md
```diff
@@ -0,0 +1,216 @@
+
+# BaseApp E-mail Templates - Django
```
nossila, line 2: Lets change this title as well

nossila (changes requested):

.github/workflows/project-workflow-old-renamed-packages.yml
Ronan-Fernandes: we still want to do a final publish on this, but now that i'm thinking i agree we don't need to run `github-actions` for `baseapp-email-templates`

baseapp-email-templates/setup.py
Ronan-Fernandes: I think we do want to require this since we want the old package to point to this new package if somene runs `pip install baseapp-email-templates`.

I took this change from this steps https://github.com/simonw/pypi-rename

pablogadhi (approved):

nossila (approved):

nossila (approved):

PR #73, opened 2024-02-02
fix: trigger push notification only if device exists

vitorguima:


ficast (approved):

baseapp-notifications/baseapp_notifications/tasks.py
```diff
@@ -13,26 +13,28 @@ def send_push_notification(
         WNSDevice,
     )
 
+    message = force_str(push_description)
```
nossila, line 16: `force_str` actually transforms `None` into `"None"` and push_notification has `None` as default

```
>>> from django.utils.encoding import force_str
>>> force_str(None)
'None'
```

nossila (approved):

PR #74, opened 2024-02-06
feat: implement a task to trim the video with a custom duration

vitorguima:


ficast (approved):

baseapp-cloudflare-stream-field/baseapp_cloudflare_stream_field/tasks.py
```diff
@@ -14,9 +15,36 @@ def refresh_from_cloudflare(content_type_pk, object_pk, attname, retries=1):
...
+                if new_video:
+                    setattr(obj, attname, new_video)
+                    delete_original_trimmed_video.apply_async(
+                        kwargs={
+                            "old_video_uid": old_video_uid,
+                            "new_video_uid": new_video["uid"],
+                        },
+                    )
+                else:
+                    setattr(obj, attname, new_value)
```
nossila, line 46: I think you need to change this if block here, seems like `setattr(obj, attname, new_value)` will never run if you don't have this settings, which would break old projects, I think what you could to do instead of `hasattr` is `getattr` for the trim setting:

```python
getattr(settings, "CLOUDFLARE_VIDEO_AUTOMATIC_TRIM", False) and hasattr(obj, "CLOUDFLARE_VIDEO_MAX_DURATION_SECONDS")
```

And have a `setattr(obj, attname, new_value)` executing as well if those settings are not set

baseapp-cloudflare-stream-field/baseapp_cloudflare_stream_field/stream.py
```diff
@@ -61,3 +65,23 @@ def download_video(self, video_uid):
...
+    def clip_video(self, data={}):
+        res = requests.post(
+            f"{self._api_url}/clip",
+            headers=self.request_headers,
+            json=data,
+        )
+
+        if res.status_code == 400:
+            logging.info(res.text)
+            logging.info(res.status_code)
```
nossila, line 78: or better format this into 1 line, maybe with `logging.error` instead of you wanna track this

baseapp-cloudflare-stream-field/baseapp_cloudflare_stream_field/stream.py
```diff
@@ -39,10 +42,11 @@ def update_video_data(self, video_uid, data={}):
         )
         return res.json()["result"]
 
-    def delete_video_data(self, video_uid):
+    def delete_video_data(self, video_uid, data={}):
         res = requests.delete(
             f"{self._api_url}/{video_uid}",
             headers=self.request_headers,
+            json=data,
```
nossila, line 49: this `data` don't seems to be used

nossila (changes requested):

nossila (approved):

PR #83, opened 2024-03-01
Update stream.py

padaliyajay:
I added method to upload caption(subtitle) file to cloudflare stream. 

User can add subtitle file field in admin form which will use to upload file which will later upload to stream from save method

### DIFF
--- a/baseapp-cloudflare-stream-field/baseapp_cloudflare_stream_field/stream.py
+++ b/baseapp-cloudflare-stream-field/baseapp_cloudflare_stream_field/stream.py
@@ -42,6 +42,14 @@ class StreamClient:
         )
         return res.json()["result"]
 
+    def upload_caption_file(self, video_uid, language_code, caption_file):
+        res = requests.put(
+            f"{self._api_url}/{video_uid}/captions/{language_code}",
+            headers={key: value for key, value in self.request_headers.items() if key != "Content-Type"},
+            files={"file": caption_file},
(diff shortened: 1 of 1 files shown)


baseapp-cloudflare-stream-field/baseapp_cloudflare_stream_field/stream.py
```diff
@@ -42,6 +42,14 @@ def update_video_data(self, video_uid, data={}):
         )
         return res.json()["result"]
 
+    def upload_caption_file(self, video_uid, language_code, caption_file):
+        res = requests.put(
+            f"{self._api_url}/{video_uid}/captions/{language_code}",
+            headers={key: value for key, value in self.request_headers.items() if key != "Content-Type"},
```
nossila, line 48: Maybe instead of looping through `self.request_headers` you could instead do:

```python
headers={
    **self.request_headers,
    "Content-Type": "multipart/form-data"
}
```

nossila (changes requested):

padaliyajay:
good idea. check again

baseapp-cloudflare-stream-field/baseapp_cloudflare_stream_field/stream.py
padaliyajay: Good idea. check now

nossila (approved):

PR #86, opened 2024-03-02
BA-1305: Inject up-to-date claims to access token on refresh

pablogadhi:
This fixes the issues with the user update flows, where the token has stale data. Whenever we update the user's data, we can just call `refreshAccessToken(ACCESS_COOKIE_NAME, REFRESH_COOKIE_NAME)` on success and that should get a new access token with the fresh user data and save it to the cookie. In my tests I also updated the useUser hook to be like the following:
```
import { useEffect, useState } from 'react'

import { getUser } from '@baseapp-frontend/authentication'

import { IUser } from 'types/user'

export const useUser = () => {
  const [user, setUser] = useState<Partial<IUser> | null>(null)

  useEffect(() => {
    const syncGetUser = async () => {
      setUser(await getUser())
    }

    syncGetUser()
  }, [])

  return user
}
```
because I wasn't sure if the getUser function can be used directly in a react hook, since it's async. Just FYI if that helps someone.

### DIFF
--- a/baseapp-auth/README.md
+++ b/baseapp-auth/README.md
@@ -89,6 +89,16 @@ SIMPLE_JWT = {
 }
 JWT_CLAIM_SERIALIZER_CLASS = "baseapp_auth.rest_framework.users.serializers.UserBaseSerializer"
 ```
+We override some settings from the Simple JWT library by default to make some baseapp features work properly. They look like the following:
+```py
+SIMPLE_JWT = {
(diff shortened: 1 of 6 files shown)


Janekk (approved):
Looks good :) maybe we can update the README of baseapp_auth with the new setting and metioned that it will be used by default.

anicioalexandre (approved):
working nice!

nossila:
@pablogadhi maybe also increase the version so we can release the update? :)

pablogadhi:
Updated the docs to talk about the default settings and bumped the package version up.

PR #104, opened 2024-04-19
Allow-for-Superuser-deletion

MattGalvis:
```
Acceptance Criteria 
    a superuser should be able to delete any other user, including a superuser
    superuser logs should not block superusers from being removed 
Current behavior 
    superusers are blocked from deletion due to superuser logs 
Expected behavior 
    a superuser should be able to delete any other user, including a superuser
    superuser logs should not block superusers from being removed
```

https://www.loom.com/share/28473927ffa745f6836e8cda95e6101b


nossila:
@MattGalvis you need to increase the version in setup.cfg in that package if you want that released in pypi soon

Janekk (approved):

nossila (changes requested):
We need to have a custom base [Event model](https://django-pghistory.readthedocs.io/en/3.2.0/settings/#pghistory_base_model) in `baseapp-core`  to not break compatibility on projects [using our fork](https://github.com/Opus10/django-pghistory/pull/69/files)

nossila (approved):

PR #116, opened 2024-05-30
BA-1247 Improve Email Verification Flow

junioresc:
Acceptance Criteria 
Context
This user story aims to improve the email authentication process. The feature ensures users verify their email addresses after registration before gaining full access. This requirement can be optionally enabled or disabled for different projects. The main flows include sending a verification email after sign-up, handling login attempts without email verification, and requiring email verification for accessing certain routes.

Scenario 1: After Sign-Up

Given

A new user registers on the application.

When

The user completes the registration and clicks the sign-up button.

Then

The system sends a verification email to the user's email address.

The system displays a screen informing the user that a verification email has been sent and that they need to verify their email to complete the registration.

trigger the email verification only once after signup

includes a button to resend the verification email

The user is not automatically authenticated after signing up.

 

Scenario 2: Login Attempt without Email Verification

Given

A user who has registered but has not yet verified their email tries to log in.

WhenThe user enters their credentials and clicks the login button.

Then

The system allows the login but does not redirect the user to the homepage or any other protected page.

The system displays a screen informing the user that a verification email has been sent and that they need to verify their email to complete the registration.

don't send the email verification again (was already sent during signup)

includes a button to resend the verification email

 

Scenario 3: Accessing Authenticated Route without Email Verification

Given

A logged-in user tries to access a page that requires email verification.

When

The user navigates to a route that requires email verification.

Then

The system displays a screen informing the user that a verification email has been sent and that they need to verify their email to complete the registration.

don't send the email verification again (was already sent during signup)

includes a button to resend the verification email

 

General Requirements

After successfully verifying the email, display the verification success screen

If the verification fails, display the verification failed screen

It must be easy to enable or disable the email verification requirement for projects that do not need this functionality.

For authenticated routes 9scenario 3), use a withAuthentication HOC to conditionally render the email non-verified page if the user's email has not been verified.

email template for the email verification (if you don't have time, use the regular template v3 already has, we can have a separate story just for applying those new templates):   

Approvd 
https://app.approvd.io/projects/BA/stories/27824


baseapp-auth/baseapp_auth/rest_framework/confirm_email/serializers.py
```diff
@@ -0,0 +1,21 @@
+from django.contrib.auth import get_user_model
+from django.utils.translation import gettext_lazy as _
+
+from rest_framework import serializers
+
+User = get_user_model()
```
deboracosilveira, line 6: looks like this model is not being used here

baseapp-auth/baseapp_auth/rest_framework/users/serializers.py
```diff
@@ -21,6 +23,7 @@
 
 class UserBaseSerializer(ModelSerializer):
     avatar = AvatarField(required=False, allow_null=True)
+    verification_required = serializers.SerializerMethodField()
```
deboracosilveira, line 26: maybe change to email_verification_required

deboracosilveira (commented):
code looks fine, but i could not test it bec it wasn't possible to checkout to your branch. I imagine that is bec you are not part of the Developers team of silverlogic organization. I think @nossila or cristina can add you

PR #118, opened 2024-06-10
BA-1247 Improve email verification flow

junioresc:
Acceptance Criteria 
Context
This user story aims to improve the email authentication process. The feature ensures users verify their email addresses after registration before gaining full access. This requirement can be optionally enabled or disabled for different projects. The main flows include sending a verification email after sign-up, handling login attempts without email verification, and requiring email verification for accessing certain routes.

Scenario 1: After Sign-Up

Given

A new user registers on the application.

When

The user completes the registration and clicks the sign-up button.

Then

The system sends a verification email to the user's email address.

The system displays a screen informing the user that a verification email has been sent and that they need to verify their email to complete the registration.

trigger the email verification only once after signup

includes a button to resend the verification email

The user is not automatically authenticated after signing up.

https://www.figma.com/design/XRD6wSl1m8Kz6XUcAy5CLp/BaseApp-Minimal-v5.4.0?node-id=331-168586&m=dev

Scenario 2: Login Attempt without Email Verification

Given

A user who has registered but has not yet verified their email tries to log in.

WhenThe user enters their credentials and clicks the login button.

Then

The system allows the login but does not redirect the user to the homepage or any other protected page.

The system displays a screen informing the user that a verification email has been sent and that they need to verify their email to complete the registration.

don't send the email verification again (was already sent during signup)

includes a button to resend the verification email

https://www.figma.com/design/XRD6wSl1m8Kz6XUcAy5CLp/BaseApp-Minimal-v5.4.0?node-id=1902-36366&m=dev

Scenario 3: Accessing Authenticated Route without Email Verification

Given

A logged-in user tries to access a page that requires email verification.

When

The user navigates to a route that requires email verification.

Then

The system displays a screen informing the user that a verification email has been sent and that they need to verify their email to complete the registration.

don't send the email verification again (was already sent during signup)

includes a button to resend the verification email

https://www.figma.com/design/XRD6wSl1m8Kz6XUcAy5CLp/BaseApp-Minimal-v5.4.0?node-id=420-227997&m=dev

General Requirements

After successfully verifying the email, display the verification success screen

If the verification fails, display the verification failed screen

It must be easy to enable or disable the email verification requirement for projects that do not need this functionality.

For authenticated routes 9scenario 3), use a withAuthentication HOC to conditionally render the email non-verified page if the user's email has not been verified.

email template for the email verification (if you don't have time, use the regular template v3 already has, we can have a separate story just for applying those new templates): https://www.figma.com/design/XRD6wSl1m8Kz6XUcAy5CLp/BaseApp-Minimal-v5.4.0?node-id=215-285093&m=dev 

Approvd 
https://app.approvd.io/projects/BA/stories/27824


nossila (approved):

deboracosilveira (approved):

anicioalexandre:
@junioresc we need to make sure we pump both packages on the `setup.cfg` file. @nossila do we have a guide to help people on how to do it?

anicioalexandre (approved):

anicioalexandre:
would be nice if we could encode/decode the ConfirmEmailTokenGenerator, something like:

```
class ConfirmEmailTokenGenerator(TokenGenerator):
    key_salt = "confirm_email"

    def get_signing_value(self, user):
        return [user.pk, user.email]

    def make_token(self, obj):
        encoded_token = urlsafe_base64_encode(force_bytes(super().make_token(obj)))
        return encoded_token

    def decode_token(self, token):
        try:
            decoded_token = urlsafe_base64_decode(token).decode("utf-8")
            return super().decode_token(decoded_token)
        except DjangoUnicodeDecodeError:
            return None
```

it causes annoying errors on the frontend (nextjs decodes the token automatically)
@nossila @Janekk  do you think we should have a little story to review the token generators we currently have on the BA? And make sure they all use `urlsafe_base64_decode`. Maybe do it by default on the `TokenGenerator`

anicioalexandre (approved):

Janekk:
@anicioalexandre agreed on base64 encoding the ConfirmEmailToken. Btw. how does the FE habdle it rn?

anicioalexandre:
> @anicioalexandre agreed on base64 encoding the ConfirmEmailToken. Btw. how does the FE habdle it rn?
it doesn't handle it well, the frontend ends up retrieving a bad token from the URL, if there are special characters in the token

nossila (approved):

PR #134, opened 2024-07-23
Ba 1375  backend run test migration health check on each package

sstsl:
### 1
Add test migration health check to each package like:

https://github.com/silverlogic/baseapp-backend/blob/baseapp-block/baseapp-blocks/baseapp_blocks/tests/test_migration_health_check.py

This is to unsure the packages are not making changes and forgetting to commit the migrations 

### 2
Save IP address on pghistory's context here https://github.com/silverlogic/baseapp-backend/blob/master/baseapp-core/baseapp_core/middleware.py#L73

Use of ipware package to make it easy to configure per environment: https://github.com/un33k/django-ipware


nossila:
@sstsl instead of copying the migration health check over and over, do whats was done in https://github.com/silverlogic/baseapp-backend/blob/baseapp-block/baseapp-blocks/baseapp_blocks/tests/test_migration_health_check.py which basically you import that test from baseapp_core instead, that way we don't need to duplicate code.

Also would be nice to revert the lint update and make a dedicate PR for the lint updates, it would be better for code review.

nossila (approved):

PR #138, opened 2024-08-14
baseapp-message-templates: add send_template_email util

pt-tsl:


baseapp-message-templates/baseapp_message_templates/email_utils.py
```diff
@@ -0,0 +1,28 @@
+import logging
+
+from .models import EmailTemplate
+
+logger = logging.getLogger(__name__)
+
+
+def send_template_email(
```
nick-prat, line 8: I think it would be better to have the args as
```
(template_name, *Args, **kwargs)
```
and then call the send method with
```
(*args, **kwargs)
```
so you don't have to manually redefine the args, and default values .

anicioalexandre (approved):
Nice! Could we improve the README a little bit? Showing examples on how to send emails using the lib and also how you could set your own email template

nick-prat (approved):

baseapp-message-templates/baseapp_message_templates/email_utils.py
```diff
@@ -0,0 +1,24 @@
...
+        logger.info(f'Email template "{template_name}" not found.')
+
+
+def send_sendgrid_email(template_name, recipient_email: str, context={}):
+    try:
+        template = EmailTemplate.objects.get(name=template_name)
+        personalization = get_personalization(recipient_email, context)
+        template.send_via_sendgrid(personalization)
+    except EmailTemplate.DoesNotExist:
+        logger.info(f'Email template "{template_name}" not found.')
```
nossila, line 24:
```suggestion
        logger.error(f'Email template "{template_name}" not found.')
```
same here

nossila (changes requested):

nossila (approved):

PR #140, opened 2024-08-15
baseapp-auth: update emails

pt-tsl:


baseapp-core/baseapp_core/templates/emails/base.html.j2
anicioalexandre, line 1: Do we still need to have this file? I think we may only need it at the template level

anicioalexandre (approved):

Ronan-Fernandes (approved):

PR #144, opened 2024-08-22
baseapp-profiles: add fields biography and banner_image and add tests

nossila:
add fields biography and banner_image and add tests for delete and update mutations

baseapp-auth: add baseapp-profiles

And update other packages's testproject to use baseapp-profiles


baseapp-profiles/baseapp_profiles/graphql/object_types.py
```diff
@@ -61,14 +65,26 @@ def meta_og_image(self):
...
+class BaseProfileObjectType:
     target = graphene.Field(lambda: ProfileInterface)
     image = ThumbnailField(required=False)
+    banner_image = ThumbnailField(required=False)
 
     class Meta:
         interfaces = interfaces
         model = Profile
-        fields = ("pk", "name", "image", "target", "status", "owner", "created", "modified")
+        fields = (
```
vitorguima, line 76: should we use `__all__` instead of specifying the fields?

baseapp-auth/baseapp_auth/rest_framework/register/serializers.py
```diff
@@ -9,8 +9,7 @@
 
 
 class RegisterSerializer(serializers.Serializer):
-    first_name = serializers.CharField(required=False, allow_blank=True)
-    last_name = serializers.CharField(required=False, allow_blank=True)
+    name = serializers.CharField(required=False, allow_blank=True)
```
vitorguima, line 12: any reasons why we are changing it like that? Not sure about the advantages of it...

pt-tsl (approved):

baseapp-profiles/baseapp_profiles/signals.py
```diff
@@ -1,39 +1,28 @@
...
 
-if apps.is_installed("avatar"):
-    from avatar.models import Avatar
-else:
-    Avatar = None
-
 
 def update_user_profile(instance, **kwargs):
+    if not hasattr(instance, "profile"):
+        return
```
vitorguima, line 7: could we raise an exception instead?

maybe some

`except AttributeError`

Ronan-Fernandes (approved):

baseapp-profiles/baseapp_profiles/signals.py
nossila: @vitorguima I can't remember why I added this condition, can you believe? I will just remove, `instance.profile` is not even used in that function.

baseapp-auth/baseapp_auth/rest_framework/register/serializers.py
nossila: @vitorguima we are changing the frontend to only ask for "name" instead of both first and last, I don't want to change the user model now so it is a step before actually changing the user model as well

PR #157, opened 2024-09-26
baseapp-chats: initial package and other small fixes

nossila:


baseapp-chats/baseapp_chats/graphql/interfaces.py
```diff
@@ -0,0 +1,46 @@
...
+    def resolve_chat_rooms(self, info, **kwargs):
+        if not info.context.user.has_perm("baseapp_chats.list_chatrooms", self):
+            return ChatRoom.objects.none()
+
+        qs = ChatRoom.objects.filter(
+            participants__profile_id=self.pk,
+        ).order_by("-last_message_time", "-created")
+
+        # Exclude rooms where the profile is blocked
+        blocking_profile_ids = Block.objects.filter(actor_id=self.pk).values_list(
```
Ronan-Fernandes, line 26: is self.pk a profile pk ?

vitorguima: yeah it should be. Actually this could be different instances depending on the [object type](https://docs.graphene-python.org/en/latest/types/objecttypes/)  in which `ChatRoomsInterface` has been added to the class meta [interfaces](https://docs.graphene-python.org/en/latest/types/interfaces/)

baseapp-chats/baseapp_chats/graphql/interfaces.py
```diff
@@ -0,0 +1,46 @@
...
+
+    def resolve_chat_rooms(self, info, **kwargs):
+        if not info.context.user.has_perm("baseapp_chats.list_chatrooms", self):
+            return ChatRoom.objects.none()
+
+        qs = ChatRoom.objects.filter(
+            participants__profile_id=self.pk,
+        ).order_by("-last_message_time", "-created")
+
+        # Exclude rooms where the profile is blocked
```
vitorguima, line 25: Suggestion: `Exclude rooms blocked by the profile` and then `Exclude rooms where the profile is blocked`. Maybe change this single comment into two?

baseapp-chats/baseapp_chats/graphql/object_types.py
```diff
@@ -0,0 +1,111 @@
...
+        try:
+            room = cls._meta.model.objects.get(id=id)
+            if not info.context.user.has_perm("baseapp_chats.view_chatroom", room):
+                return None
+            return room
+
+        except cls._meta.model.DoesNotExist:
+            return None
+
+    # @classmethod
```
vitorguima, line 84: I guess we can remove it right?

vitorguima (approved):

baseapp-chats/baseapp_chats/graphql/object_types.py
```diff
@@ -0,0 +1,111 @@
...
+    # def get_queryset(cls, queryset, info):
```
vitorguima, line 85: guess we can remove it

baseapp-chats/baseapp_chats/graphql/interfaces.py
vitorguima, line 25:
```suggestion
        # Exclude rooms blocked by the profile
```

baseapp-chats/baseapp_chats/graphql/interfaces.py
```diff
@@ -0,0 +1,46 @@
...
+        blocking_profile_ids = Block.objects.filter(actor_id=self.pk).values_list(
+            "target_id", flat=True
+        )
+        qs = qs.exclude(participants__profile_id__in=blocking_profile_ids)
+
+        blocker_profile_ids = Block.objects.filter(target_id=self.pk).values_list(
```
vitorguima, line 31:
```suggestion
        ## exclude rooms where the profile is blocked
        blocker_profile_ids = Block.objects.filter(target_id=self.pk).values_list(
```

vitorguima (approved):

baseapp-chats/baseapp_chats/graphql/interfaces.py
nossila: on this implementation only the Profile can implement `ChatRoomsInterface`

PR #167, opened 2024-10-07
baseapp-profiles: enable image deletion on update

pt-tsl:
Enable removing images when updating profiles from `baseapp-profiles`


nossila:
Can we include some tests to cover this cases? You can find some tests related to image and banner_image here https://github.com/silverlogic/baseapp-backend/blob/d7111e34d73fb936edcb205def930bb960b56429/baseapp-profiles/baseapp_profiles/tests/test_graphql_mutations_update.py#L103

anicioalexandre (approved):

nossila (approved):

PR #170, opened 2024-10-15
baseapp-comments-fix-migration-for-swapping-model

vitorguima:


procopio420 (approved):

baseapp-comments/baseapp_comments/migrations/0001_initial.py
```diff
@@ -11,6 +11,7 @@
 
 import baseapp_comments.models
 import baseapp_comments.validators
+import swapper
```
nick-prat, line 14: What happens if you change the model after these migrations have been run?

vitorguima: u mean after installing the package and customizing it? it will create new migrations in your project.

nick-prat: What I mean is, lets say you started off the project with the baseapp comments model, and then all these migrations have been applied with said model.  After you switch the model, these migrations effectively change, but they won't be re-run, as django will think they've been applied already.

vitorguima: yeah that's a good catch @nick-prat and that's why I had to do these changes in order to solve come conflicts for this scenario. Due to some circular dependencies that were happening and conflicts due to inexistent fields applied in the old migrations.

but right now the way it is, when customizing the model after the initial setup, u'll need to:

- fake the migration from the package to the initial statue `./manage.py migrate baseapp_package_name zero --fake`
- then run the migrate again with the swappable model.

Me and @nossila talked about it when I was doing this story to maybe in the future we just work with inheritance instead of using swappable models. But I don't think we are going to change in the short term. So yeah not that pretty though.

Ronan-Fernandes (approved):

nossila (approved):

nossila (approved):

PR #172, opened 2024-10-29
Fix remove superuser e-mail

rafaelhdr:
Fix 500 error

```
TemplateDoesNotExist at /admin/users/user/2/change/

users/emails/remove-superuser-notification-email.txt.j2
```

And this also includes a test to detect the same error.

### DIFF
--- a/baseapp-auth/baseapp_auth/tests/unit/test_emails.py
+++ b/baseapp-auth/baseapp_auth/tests/unit/test_emails.py
@@ -1,0 +1,17 @@
+import baseapp_auth.tests.helpers as h
+import pytest
+
+from ...emails import remove_superuser_notification_email
+
+UserFactory = h.get_user_factory()
+
+pytestmark = pytest.mark.django_db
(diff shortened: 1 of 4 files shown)


baseapp-auth/baseapp_auth/tests/unit/test_emails.py
```diff
@@ -0,0 +1,17 @@
...
+
+from ...emails import remove_superuser_notification_email
+
+UserFactory = h.get_user_factory()
+
+pytestmark = pytest.mark.django_db
+
+
+def test_remove_superuser_notification_email(outbox):
+    UserFactory(is_superuser=True)
```
rafaelhdr, line 12: This line seems not useful for the tests, but it is actually necessary for having a recipient_list in the e-mail, and then we will get an e-mail detected in `outbox`

nossila (approved):

vitorguima (approved):

PR #187, opened 2024-11-22
baseapp-message-templates: translations for email templates

pt-tsl:
Add a model to enable translations for email templates

## Release Notes

- **New Features**
	- Introduced support for template translations in multiple languages (English, Spanish, Portuguese).
	- Added a new inline admin interface for managing template translations alongside email attachments.
	- Enhanced email sending functionality to prioritize language-specific content.
	- Updated the email template management to support translations with a new `TemplateTranslation` model.

- **Bug Fixes**
	- Improved error handling for sending emails without translations.

- **Documentation**
	- Updated settings to include new language options for internationalization.

- **Tests**
	- Expanded test coverage for email and SMS templates, including translation scenarios.

- **Chores**
	- Incremented package version from 1.9.3 to 2.0.0 to reflect significant updates.


anicioalexandre (approved):

Ronan-Fernandes (approved):

baseapp-message-templates/baseapp_message_templates/models.py
```diff
@@ -18,21 +21,16 @@ def db_type(self, connection):
...
+    target_content_type = models.ForeignKey(
+        ContentType,
         blank=True,
         null=True,
+        on_delete=models.CASCADE,
     )
-
-    # Core
+    target_object_id = models.PositiveIntegerField(blank=True, null=True)
+    target = GenericForeignKey("target_content_type", "target_object_id")
```
nick-prat, line 32: Why are we using generic foreign keys here?  Are these going to provide translations for something other than EmailTemplate?

nick-prat (commented):
I thought the plan was to make it so that the email translations would provide a translated context that would be used to set the text, instead of redefining the HTML for each translation?

As it stands, if we're going to redefine the HTML structure as well as the text for each template, then I'm not sure if we would need the TemplateTranslation at all, since we could just create an EmailTemplate for each language, and added a 'language' column on it.

PR #197, opened 2024-12-10
baseapp-profiles: ProfilableModel has protected instead of Cascade

rafaelhdr:
It is not something that happen much, but if we try to delete a profile, we should protect the user (instead of CASCADE). In my example, it was for tests we were doing on staging.

As result, Django Admin should project w/ `Cannot delete profile` instead of deleting also users.

## Summary by CodeRabbit

- **New Features**
	- Enhanced deletion logic for profiles to ensure they cannot be deleted if still associated with a user.
- **Bug Fixes**
	- Updated the `on_delete` behavior for the profile field to prevent accidental deletions.
- **Migrations**
	- Modified the `profile` field in the user model to establish a one-to-one relationship, allowing for optional profiles and improving usability.
	- Added new middleware to handle messages and notifications within the application.


nossila (changes requested):
You will need to makemigrations on all testproject on the other packages as well. Tests are failing mostly because of that missing migrations.

baseapp-profiles/baseapp_profiles/models.py
```diff
@@ -145,7 +145,7 @@ class ProfilableModel(models.Model):
     profile = models.OneToOneField(
         swapper.get_model_name("baseapp_profiles", "Profile"),
         related_name="%(class)s",
-        on_delete=models.CASCADE,
+        on_delete=models.PROTECT,
         verbose_name=_("profile"),
         null=True,
         blank=True,
```
nick-prat, line 151: If we're supporting a `ProfilableModel` to _not_ have a profile, why are we not letting them delete said profile once it's been created?  I'm usually suspicious of protect delete behavior on a nullable foreign key, would `SET_NULL` not be more appropriate?

rafaelhdr: Ooooh I thought we wanted all users to have profile. If not, I agree with you.

Up to you @nossila @nick-prat @vitorguima decide what is better for us.

nossila: @nick-prat its nullable because we were working on projects that already had Users, we could remove the null now if desired.

@rafaelhdr are you looking into having users with profile and without profile mixed up in the same project? I also think that `SET_NULL` might be more appropriate in that case. But the initial idea was to really have a profile for every user, in case that `CASCADE` is fine and easier to handle deletes, but I also like the idea of `PROTECT` that would make you always delete the users and not the profile. We can even remove DELETE option on django admin for Profiles. What you think?

vitorguima: @nick-prat @nossila since we need a definition for this let's go with the `SET_NULL` then? So we don't touch the nullable logic and we can keep the package more flexible. I also don't see any big counter points for not using `SET_NULL`.

rafaelhdr: > are you looking into having users with profile and without profile mixed up in the same project?
Nope. But I think you know the roadmap better than me

> we could remove the null now if desired.
@nossila can we remove the nullable?
I think we would have the egg and chicken issue here, no?
We can't create the `User`, because we don't have the `Profile` yet. And we can't create the `Profile` because we don't have the `User` yet.

So, for now, everyone agrees with `SET_NULL`? And then later we discuss better what would be better?

rafaelhdr: > We can even remove DELETE option on django admin for Profiles.
I think regardless of the decision, this seems a good idea. I believe someone was doing something wrong (to delete an user profile) and this Pull Request just solves the error 500. If we don't want people deleting the Profile, it seems a good idea for me.

nick-prat (approved):

vitorguima (approved):

baseapp-profiles/baseapp_profiles/models.py
rafaelhdr: I will resolve the conversation, but just for transparency.

We discussed in Slack and decided to keep w/ PROTECTED. SET_NULL makes sense, but we don't want to have users without Profile. And this is very configurable. If any project want a different behavior, can set directly in the template.

nossila:
@rafaelhdr, @nick-prat and @vitorguima I change the implementation to set `User.profile` to `SET_NULL` but any other `ProfilableModel.profile` to `PROTECT` by default.

Because of circular foreign key between `User.profile` and `Profile.owner` it made it impossible to delete either one or even truncate the table, making `baseapp-e2e` [tests fail](https://github.com/silverlogic/baseapp-backend/actions/runs/14244660533/job/39922460243#step:10:141).

I think it make sense to allow Users with no Profile, with that we can have some users as admin only for example.

What are your thoughts on this?

nick-prat:
Makes sense to me @nossila

vitorguima:
> @rafaelhdr, @nick-prat and @vitorguima I change the implementation to set `User.profile` to `SET_NULL` but any other `ProfilableModel.profile` to `PROTECT` by default.
> 
> Because of circular foreign key between `User.profile` and `Profile.owner` it made it impossible to delete either one or even truncate the table, making `baseapp-e2e` [tests fail](https://github.com/silverlogic/baseapp-backend/actions/runs/14244660533/job/39922460243#step:10:141).
> 
> I think it make sense to allow Users with no Profile, with that we can have some users as admin only for example.
> 
> What are your thoughts on this?

Makes sense! Sounds good to me.

PR #198, opened 2024-12-10
baseapp-chats: create chat group

Ronan-Fernandes:
## Summary by CodeRabbit

- **New Features**
	- Introduced a new field to differentiate between group and individual chat rooms.
	- Added optional fields for creating chat rooms, including `is_group` and `title`.

- **Bug Fixes**
	- Implemented validation to ensure a title is provided for group chats.

- **Tests**
	- Enhanced GraphQL mutation tests for group chat creation with new test cases.

- **Chores**
	- Updated version number of the application to 0.0.7.

PR #208, opened 2024-12-16
feat: filter past messages for new chat group participant

pt-tsl:
**Acceptance Criteria** 
**Display Messages based on time Participant Joined**

- Given I am a new participant added to a group, when I open the group chat, then I should be able to see messages from the time I was added to the group, but not earlier.
- If no messages have been sent, we should displayed "You were added to this group".


baseapp-chats/baseapp_chats/graphql/object_types.py
```diff
@@ -98,6 +98,19 @@ class BaseChatRoomObjectType:
     image = ThumbnailField(required=False)
 
     def resolve_all_messages(self, info, **kwargs):
+        if self.participants_count > 2:
```
nossila, line 101: since we are introducing `is_group` in [another PR](https://github.com/silverlogic/baseapp-backend/pull/198/files) we should base on that instead of `participants_count`

nossila (changes requested):

nossila (approved):

anicioalexandre (approved):

PR #215, opened 2025-01-07
baseapp-chats: Removing chat room participants

tsl-ps2:
As a user, on the BaseApp Messages, I would like to Remove an existing member of the group, In order to avoid that person to have access to the group messages.

Original Story:  
**Acceptance Criteria 
Removing a Member**

- Given I am an admin of the group, when I select a member from the group details view, then I should have the option to remove that member from the group.

  - We should have already implemented the Remove Member option button in the Admin Permission story, but not the functionality. 

- Given I click on remove member, I should see a confirmation dialog, to make sure I want to do this action.

- Given I confirm the removal of the member, then the removed member should no longer see the group in their chat list or have access to its messages.

  - We are doing a hard delete of the group chat for that user. 

## Summary by CodeRabbit

- **New Features**
	- Enhanced chat room participant management with ability to remove participants
	- Added participant role tracking in chat rooms
	- Improved chat room update notifications

- **Bug Fixes**
	- Streamlined participant removal process
	- Enhanced error handling for chat room operations

- **Documentation**
	- Updated GraphQL schema to provide more detailed chat room information

- **Chores**
	- Incremented package version to 0.0.11

### DIFF
--- a/baseapp-chats/baseapp_chats/tests/test_graphql_mutations.py
+++ b/baseapp-chats/baseapp_chats/tests/test_graphql_mutations.py
@@ -45,6 +45,7 @@ CREATE_ROOM_GRAPHQL = """
                     id
                     title
                     isGroup
+                    participantsCount
                     participants {
                         edges {
                             node {
@@ -73,16 +74,26 @@ UPDATE_ROOM_GRAPHQL = """
                     id
                     title
                     isGroup
+                    participantsCount
                     participants {
                         edges {
                             node {
                                 id
+                                profile {
+                                    id
+                                }
                             }
                         }
                     }
                     image(width: 100, height: 100) {
                         url
                     }
+                }
+            }
+            removedParticipants {
+                id
+                profile {
+                    id
                 }
             }
             errors {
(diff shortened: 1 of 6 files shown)


baseapp-chats/baseapp_chats/tests/test_graphql_mutations.py
```diff
@@ -85,6 +90,13 @@
                     }
                 }
             }
+            removedParticipants {
```
nossila, line 93: usually for the frontend only needs the IDs for the node/objects that got removed, but for me its fine to leave as it is

nossila (approved):

baseapp-chats/baseapp_chats/tests/test_graphql_mutations.py
tsl-ps2: I will delete the `name` from profile, it is indeed not needed. (I need the `profile` with its `id` though, just the general id of the removed participant is not enough.)

Ronan-Fernandes (approved):

PR #219, opened 2025-01-15
baseapp-chats: add rich text field to messages

Ronan-Fernandes:
## Release Notes for baseapp-chats v0.0.10

- **New Features**
  - Added support for rich text content in messages.
  - Introduced plain text extraction for message content.
  - Enhanced HTML sanitization for message inputs.

- **Improvements**
  - Improved message content handling with richer text formatting.
  - Added content sanitization to prevent potential security risks.

- **Chores**
  - Updated package version to 0.0.10.


baseapp-chats/baseapp_chats/graphql/mutations.py
```diff
@@ -363,7 +369,7 @@ def mutate_and_get_payload(
                 ]
             )
 
-        if len(content) > 1000:
+        if len(clean_content) > 1000:
```
tsl-ps2, line 372: Just to make sure: I would expect that sanitizing content only decreases its length? Otherwise, it would make for a bad user experience if content with length <= 1000 on the client became longer and fails this validation.

tsl-ps2 (approved):
Left one comment.

baseapp-chats/baseapp_chats/base.py
```diff
@@ -53,7 +55,7 @@ class Verbs(models.IntegerChoices):
     def description(self):
         return self.label
 
-    content = models.TextField(null=True, blank=True)
+    content = RichTextField(blank=True, null=True)
```
nossila, line 58: RichTextField worries me a lot, it means we are dealing with raw HTML and that can come with major security implications, I would look into using a more simple formatting language like markdown for example, its a big reason to why github comments uses it for example

nossila (changes requested):

PR #220, opened 2025-01-15
baseapp-chats [BA-2108]: leave group as admin

pt-tsl:
**Acceptance Criteria**

- Given I click on the "Leave Group" button in the 3 dots menu, when the action is triggered, then I should see a confirmation prompt asking if I am sure I want to leave the group.

- Given I am the sole admin in the group chat, then the confirmation dialog should inform me prompt me to select a new admin before leaving and informing me if I don't, then the most senior member will be the new admin.

- Given there are other admins besides me in the group then only display a the same confirmation dialog a normal member would have when leaving the group chat.

- Given I confirm that I want to leave the group, then I should no longer see the group in my chat list and stop receiving updates of messages on that group.

We are going to do a hard delete for that Group Chat for that user.

- Given I've left the chat room, then make the most senior member in the group the new admin.

### DIFF
--- a/baseapp-chats/baseapp_chats/graphql/mutations.py
+++ b/baseapp-chats/baseapp_chats/graphql/mutations.py
@@ -222,6 +222,31 @@ class ChatRoomUpdate(RelayMutation):
     ):
         room = get_obj_from_relay_id(info, room_id)
         profile = get_obj_from_relay_id(info, profile_id)
+        is_sole_admin = (
+            ChatRoomParticipant.objects.filter(
+                profile=profile, room=room, role=ChatRoomParticipantRoles.ADMIN
+            ).exists()
+            and ChatRoomParticipant.objects.filter(
+                room=room, role=ChatRoomParticipantRoles.ADMIN
+            ).count()
+            == 1
+        )
+        is_leaving_chatroom = profile_id in remove_participants and len(remove_participants) == 1
+
+        add_participants_pks = [
+            get_pk_from_relay_id(participant) for participant in add_participants
+        ]
+
+        remove_participants_pks = [
+            get_pk_from_relay_id(participant) for participant in remove_participants
+        ]
+        participants_to_remove = ChatRoomParticipant.objects.filter(
+            profile_id__in=remove_participants_pks, room=room
+        )
+        removed_participants = list(participants_to_remove)
+
+        title = input.get("title", None)
+        image = info.context.FILES.get("image", None)
 
         if not room or not getattr(room, "is_group", False):
             return ChatRoomUpdate(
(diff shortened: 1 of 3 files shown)


Ronan-Fernandes (approved):

baseapp-chats/baseapp_chats/graphql/object_types.py
```diff
@@ -139,6 +140,24 @@ def resolve_all_messages(self, info, **kwargs):
                 )
         return self.messages.all().order_by("-created")
 
+    def resolve_participants(self, info, **kwargs):
+        if self.is_group:
+            profile = (
+                info.context.user.current_profile
```
tsl-ps2, line 146: `info.content.user.current_profile` looks like a profile, but `info.context.user.profile.pk` (the value in the else statement) is a primary key. Is this intended?

baseapp-chats/baseapp_chats/permissions.py
```diff
@@ -55,11 +55,12 @@ def can_modify_chatroom(self, user_obj, obj):
             if not chat_room_participant:
                 return False
 
-            is_removing_self = len(remove_participants) == 1 and str(remove_participants[0]) == str(
-                current_profile.pk
-            )
-
-            if is_removing_self and not modify_image and not modify_title and not add_participants:
+            if (
```
tsl-ps2, line 58: Do we also need to check that no other participant is removed? Otherwise it looks like even if I do not have admin permissions, I can still remove people from a chatroom if I leave at the same time.

baseapp-chats/baseapp_chats/graphql/mutations.py
```diff
@@ -277,37 +296,34 @@ def mutate_and_get_payload(
...
-        )
-        removed_participants = list(participants_to_remove)
         participants_to_remove.delete()
 
+        # Setting new admin if needed
+        if is_leaving_chatroom and is_sole_admin:
+            oldest_remaining_participant = (
+                ChatRoomParticipant.objects.filter(room=room).order_by("accepted_at").first()
+            )
+            oldest_remaining_participant.role = ChatRoomParticipantRoles.ADMIN
```
tsl-ps2, line 313: Should we check here that `oldest_remaining_participant` is not `None` (that the leaving person is not the last chat room participant)?

tsl-ps2 (changes requested):
I left 3 comments.

baseapp-chats/baseapp_chats/graphql/mutations.py
pt-tsl: I think this is something that needs to be addressed with the team, as we never talked about what happens with a group chat if the number of members falls below 3.

nick-prat (approved):

PR #221, opened 2025-01-16
baseapp-chats [BA-2047]: leave group

pt-tsl:
**Acceptance Criteria** 
_Leave group - As a Member_

- Given I am a member of a Group shat, I should see a 3 dots menu in the Chat room title area.
- Given I am a member of a group chat, when click the 3 dots menu in the Group Title, then I should see a "Leave Group" option prominently displayed.
- Given I click on the "Leave Group" button, when the action is triggered, then I should see a confirmation prompt asking if I am sure I want to leave the group.
- Given I confirm that I want to leave the group, then I should no longer see the group in my chat list and stop receiving updates of messages on that group.

**We are going to do a hard delete for that Group Chat for that user.**

### DIFF
--- a/baseapp-chats/baseapp_chats/graphql/mutations.py
+++ b/baseapp-chats/baseapp_chats/graphql/mutations.py
@@ -243,18 +243,25 @@ class ChatRoomUpdate(RelayMutation):
                 ]
             )
 
-        add_participants = [
-            get_obj_from_relay_id(info, participant) for participant in add_participants
+        add_participants_pks = [
+            get_pk_from_relay_id(participant) for participant in add_participants
         ]
-        add_participants = [
-            participant for participant in add_participants if participant is not None
+
+        remove_participants_pks = [
+            get_pk_from_relay_id(participant) for participant in remove_participants
         ]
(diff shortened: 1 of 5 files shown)


Ronan-Fernandes (approved):
Nice, maybe we should make a test for this new functionality

tsl-ps2 (approved):
Looks good to me! If you wanted to, you could maybe add a test case that users have the permission to leave a room if they do not update any of the other settings. (Edit: just saw that Ronan already left the same comment.)

baseapp-chats/baseapp_chats/graphql/mutations.py
```diff
@@ -278,17 +291,22 @@ def mutate_and_get_payload(
...
             return ChatRoomUpdate(
                 errors=[ErrorType(field="image", messages=serializer.errors["image"])]
             )
 
+        if image is not None:
+            room.image = serializer.validated_data["image"]
+        elif delete_image:
+            room.image = None
+
         remove_participants_ids = [
```
nick-prat, line 310: I think we define this twice now, maybe remove this one?

nick-prat (approved):

PR #222, opened 2025-01-16
baseapp-auth[allauth]: Add user device app and implement user agent info

sundayguru:


nossila (approved):

baseapp-auth/baseapp_devices/models.py
```diff
@@ -0,0 +1,29 @@
...
+class BaseUserDevice(models.Model):
+    class Meta:
+        abstract = True
+
+    user = models.ForeignKey(
+        settings.AUTH_USER_MODEL,
+        related_name="devices",
+        on_delete=models.CASCADE,
+    )
+    ip_address = models.CharField(max_length=255)
```
nossila, line 15: What if we change to [GenericIPAddressField](https://docs.djangoproject.com/en/5.1/ref/models/fields/#genericipaddressfield) instead?

baseapp-auth/baseapp_devices/utils.py
```diff
@@ -0,0 +1,74 @@
+import sys
+from hashlib import md5
+
+from django.conf import settings
+from django.core.cache import DEFAULT_CACHE_ALIAS
+from user_agents import parse
+
+
+# `get_cache` function has been deprecated since Django 1.7 in favor of `caches`.
```
nossila, line 9: we only support django 4 or newer and python 3 or newer, should we care about this? the whole implementation is following old django style and supporting python 2

baseapp-auth/baseapp_devices/graphql/queries.py
```diff
@@ -0,0 +1,11 @@
+import graphene
+
+from baseapp_devices.graphql.object_types import UserDeviceType
+from baseapp_devices.models import UserDevice
+
+
+class UserDeviceQueries:
+    all_user_devices = graphene.List(UserDeviceType)
```
nossila, line 8: lets use `DjangoConnectionField` instead to make this paginated.

and maybe (just a thought/suggestion) we can add this to the `UserObjectType` instead of having it as a query, this way you could query for it like this:
```
{
   me {
      devices {
         ....
      }
  }
}
```
i think it follows more the idea of a "graph"

nossila (changes requested):

nick-prat:
Why are we reordering all the imports here, is that intentional?

sundayguru:
> Why are we reordering all the imports here, is that intentional?

Yes, as the lint where complaining.

sundayguru:
Since this is going into feature branch, I would merge it.

PR #224, opened 2025-01-17
baseapp-chats: System messages

tsl-ps2:
This implements system messages for creating a group chat. **The story below is slightly different: it asks for creating such messages when someone is added to a group. However, since this functionality does not exist yet, and all messages are kind of equivalent, this PR introduces them for creating a group chat. This was approved by FS.**

**Description**

- As a user, on the BaseApp Messages,I would like to See a message when a new member has joined the group, In order to be aware of any changes to the group membership.

**Acceptance Criteria 
Business Rules**

- Given I am a user in a group chat, when an admin has added a new member to the group, then I should see a message communicating that "[Member Name] was added".

- The user should be able to distinguish from the conversation bubbles and this system message

- The system-generated message must adhere to the chronological order of the chat history.

  - It should appear immediately after the last message sent in the chat timeline, based on the timestamp of the event when the new member joined.

- All the members in the chat group should see this message, including the person that was added. 

**Notes** 
Next story make it about the other system messages:

- Removed a member

- Left the group

- Create the group chat

- Edited name

- Edited picture

- Make someone an admin

## Summary by CodeRabbit

- **New Features**
	- Added message type categorization (user vs. system-generated messages)
	- Introduced ability to link messages to specific profiles
	- Enhanced chat room creation with automatic server-generated messages
	- Expanded GraphQL schema to include new message type and linked profile fields
	- Added `messageType` field to GraphQL query for retrieving message details

- **Bug Fixes**
	- Improved message status handling for different message types
	- Ensured unread message count is not incremented for system-generated messages

- **Tests**
	- Added test coverage for new message type and profile linking functionality
	- Verified server message behavior during chat room creation
	- Confirmed no message status is created for system-generated messages


baseapp-chats/baseapp_chats/graphql/mutations.py
```diff
@@ -168,6 +168,16 @@ def mutate_and_get_payload(cls, root, info, profile_id, participants, is_group,
         room.participants_count = len(created_participants)
         room.save()
 
+        if is_group:
+            send_message(
+                message_type=Message.MessageType.SYSTEM_GENERATED,
+                room=room,
+                profile=None,
+                user=None,
+                content='{content_linked_profile} created group "' + title + '"',
```
anicioalexandre, line 177: lets put that on a constant, we'll probably need when sending new system messages

baseapp-chats/baseapp_chats/base.py
```diff
@@ -49,11 +49,22 @@ class AbstractBaseMessage(TimeStampedModel, RelayModel):
...
+    class MessageType(models.IntegerChoices):
+        USER_MESSAGE = 1, _("User message")
+        SYSTEM_GENERATED = 2, _("System message")
+
     @property
     def description(self):
         return self.label
 
     content = models.TextField(null=True, blank=True)
+    content_linked_profile = models.ForeignKey(
```
anicioalexandre, line 61: I'm wondering what we should do when we need to send a message that needs to tie 2 or more profiles, like "Ana added you." or "Ana removed Alberto.".  Just added another FK field?

baseapp-chats/baseapp_chats/graphql/mutations.py
anicioalexandre, line 177: @tsl-ps2 @nossila can we generate the parsing on the backend instead of parsing it on the frontend? I _think_ we would have everything we need (like access to the current profile via `info.context.user.current_profile`), don't we?

baseapp-chats/baseapp_chats/base.py
tsl-ps2: Yes, adding another FK field is what I had in mind. I just added it.

baseapp-chats/baseapp_chats/graphql/object_types.py
```diff
@@ -74,27 +77,58 @@ def get_node(cls, info, id):
...
+    def get_replaced_profile_name(profile, profile_pk, replacement_text):
+        if not profile:
+            return None
+        elif profile.id == profile_pk:
+            return replacement_text
+        else:
+            return profile.name
+
+    @staticmethod
+    def resolve_content(root, info, profile_id=None, **kwargs):
```
anicioalexandre, line 107: hmmm do we need to do this in a resolve? Since it would only happen when we send a system message, maybe we could use a util function when we send the message, just to avoid running a resolver every time, something like:

```
def send_system_message(
    content,
    room,
    user_profile_pk,
    content_linked_profile_actor=None,
    content_linked_profile_receiver=None,
    self_actor_name="You",
    self_receiver_name="you",
):
    actor_name = get_replaced_profile_name(  # maybe this could be just another util function?
        content_linked_profile_actor, user_profile_pk, self_actor_name
    )
    receiver_name = get_replaced_profile_name(
        content_linked_profile_receiver, user_profile_pk, self_receiver_name
    )

    formatted_content = content.format(
        content_linked_profile_actor=actor_name,
        content_linked_profile_receiver=receiver_name,
    )

    send_message(
        message_type=Message.MessageType.SYSTEM_GENERATED,
        room=room,
        profile=None,
        user=None,
        content=formatted_content,
        content_linked_profile_capital=content_linked_profile_actor,
        content_linked_profile_capital=content_linked_profile_receiver,
    )

# usage:
send_system_message(
    content=_("{content_linked_profile_actor} has left the chat."),
    room=room,
    user_profile_pk=get_profile_pk(...),
    content_linked_profile_actor=content_linked_profile_actor,
    content_linked_profile_receiver=None,
)
```

anicioalexandre (changes requested):

baseapp-chats/baseapp_chats/graphql/object_types.py
tsl-ps2: @anicioalexandre: I am not sure whether I completely understood what you are saying. The system message is displayed differently for different people. Say A, B, C are in a chat and A adds D. Then the message displayed to A is "You added D", B and C see "A added D", and D sees "A added you". The person requesting the content should only be available in the graphql resolvers, but not when the message is generated (i.e. the `user_profile_pk` in your code snippet should be the `pk` of the person making the request). So we need a customized resolver in any case (unless we do the parsing on the FE, of course), don't we?

We could of course store the different replaced messages in the database to avoid reparsing them with every request, but then we still need a custom resolver to get the correct message for each participant and the data stored would be far from 'minimal'. So I think I would rather avoid that.

anicioalexandre: ohhh true I forgot it should be dynamic, my bad. I think we could keep the way it is, looks fine to me

baseapp-chats/baseapp_chats/graphql/mutations.py
```diff
@@ -168,6 +168,16 @@ def mutate_and_get_payload(cls, root, info, profile_id, participants, is_group,
         room.participants_count = len(created_participants)
         room.save()
 
+        if is_group:
+            send_message(
+                message_type=Message.MessageType.SYSTEM_GENERATED,
+                room=room,
+                profile=None,
+                user=None,
+                content='{content_linked_profile_actor} created group "' + title + '"',
```
anicioalexandre, line 177: maybe create {content_linked_profile_actor} and {content_linked_profile_receiver} constants? we may need to use it a lot

anicioalexandre (approved):

vitorguima (approved):

baseapp-chats/baseapp_chats/base.py
```diff
@@ -49,11 +49,29 @@ class AbstractBaseMessage(TimeStampedModel, RelayModel):
...
 
     content = models.TextField(null=True, blank=True)
+    content_linked_profile_actor = models.ForeignKey(
+        swapper.get_model_name("baseapp_profiles", "Profile"),
+        related_name="linked_as_content_actor_set",
+        on_delete=models.CASCADE,
+        null=True,
+        blank=True,
+    )
+    content_linked_profile_receiver = models.ForeignKey(
```
vitorguima, line 68: @tsl-ps2  What's the purpose of the `content_linked_profile_receiver`? Are we really using it?

baseapp-chats/baseapp_chats/graphql/mutations.py
vitorguima: @anicioalexandre so far the content will serve only for system messages which is now just being used for a new room with multiple people, right? So what's the purpose of having the `content_linked_profile_receiver`?  Maybe I'm missing something

baseapp-chats/baseapp_chats/base.py
tsl-ps2: Not yet. It is for messages like `{content_linked_profile_actor} added {content_linked_profile_receiver}`

vitorguima: I'd rather have something like `content_linked_profile_target` instead of receiver. For me it looks like it's the one who received the messaged but actually it's going to be the target of the action, right?

baseapp-chats/baseapp_chats/utils.py
```diff
@@ -27,7 +31,10 @@ def send_message(
         user=user,
         profile=profile,
         content=content,
+        content_linked_profile_actor=content_linked_profile_actor,
+        content_linked_profile_receiver=content_linked_profile_receiver,
```
vitorguima, line 35: change to `content_linked_profile_target`

vitorguima (changes requested):

baseapp-chats/baseapp_chats/graphql/mutations.py
anicioalexandre: @vitorguima  in the near future, we will have other kinds of system messages, like `"{content_linked_profile_actor} added {content_linked_profile_receiver}"`

baseapp-chats/baseapp_chats/graphql/object_types.py
```diff
@@ -74,27 +77,58 @@ def get_node(cls, info, id):
...
+        )
 
-        message_status = self.statuses.filter(profile_id=profile_pk).first()
+    @staticmethod
+    def resolve_is_read(root, info, profile_id=None, **kwargs):
+        profile_pk = BaseMessageObjectType.get_profile_pk(info, profile_id)
+        if not profile_pk:
+            return None
 
+        message_status = root.statuses.filter(profile_id=profile_pk).first()
```
nick-prat, line 131: I think this will result in an extra query for every message we return

baseapp-chats/baseapp_chats/graphql/object_types.py
```diff
@@ -74,27 +77,58 @@ def get_node(cls, info, id):
...
             return None
 
-    def resolve_is_read(self, info, profile_id=None, **kwargs):
+    @staticmethod
+    def get_profile_pk(info, profile_id=None):
         if profile_id:
             profile_pk = get_pk_from_relay_id(profile_id)
-            profile = Profile.objects.get_if_member(pk=profile_pk, user=info.context.user)
-            if not profile:
+            if not Profile.objects.get_if_member(pk=profile_pk, user=info.context.user):
```
nick-prat, line 84: Also here, I believe we're going to be running an extra query for every message...  Maybe the best thing to do would be to add the user's profiles to the request context, so we can check that instead of the DB here, going from N+1 -> 1 queries.

nick-prat (changes requested):
If I understand correctly, we're going to be making multiple DB queries for every message returned, I think we should address that

baseapp-chats/baseapp_chats/graphql/object_types.py
tsl-ps2: I think it does, yes. But I only restructured the `resolve_is_read` method, I did not make any substantial changes (unless I misunderstood something). So it shouldn't perform worse than before.

baseapp-chats/baseapp_chats/graphql/object_types.py
tsl-ps2: @nick-prat: If I understand this correctly, the query is usually not run: On the frontend, we query `content` or `isRead` without providing a `profileId`. In this case, the database query does not get executed. Only if someone queries `content` or `isRead` providing an explicit profileId, then we query the database to check whether the user is allowed to use this id. I do not know why this option is needed, I did not see it anywhere in the FE code. Maybe it is for the django admin or nice to have for debugging, I don't know. Since this option existed for the `isRead` field, I decided to treat the `message` field similarly and also add it there.

tsl-ps2:
Unless I miss something, we only need more than one database query per message if someone queries also message status properties, like `isRead` (or makes graphql queries using the profileId explicitly, I don't know why this is in the code, see my comment above). Since the message status is stored on a different model than the message content, I think it is fair to use more than one query in this case (but if you have a concrete idea how to improve this, then please let me know!)

baseapp-chats/baseapp_chats/graphql/object_types.py
nick-prat: Okay, maybe something to consider in the future then

baseapp-chats/baseapp_chats/graphql/object_types.py
nick-prat: Ah that makes sense, I guess it allows a user to query this with a profile id other then their current profile :thinking: maybe if this is a use case that we find we start using more in the future, we can deal with this extra query then.

nick-prat (approved):

vitorguima (approved):

PR #228, opened 2025-01-20
Edit message

Ronan-Fernandes:


baseapp-chats/baseapp_chats/permissions.py
```diff
@@ -124,3 +125,9 @@ def has_perm(self, user_obj, perm, obj=None):
                 return False
 
             return room.participants.filter(profile_id__in=my_profile_ids).exists()
+
+        if perm == "baseapp_chats.change_message" and user_obj.is_authenticated:
+            if isinstance(obj, Message):
+                return obj.user_id == user_obj.id
```
tsl-ps2, line 131: Should the user or profile that wrote the message have permissions to edit it? I think most other message features use the profile (we also display profile name and picture), so should we use the profile (instead of user) here?

baseapp-chats/baseapp_chats/tests/test_graphql_mutations.py
```diff
@@ -104,6 +104,23 @@
     }
 """
 
+EDIT_MESSAGE_GRAPHQL = """
+    mutation EditMessageMutation($input: ChatRoomEditMessageInput!) {
+        chatRoomEditMessage(input: $input) {
+            message {
+                node {
+                    id
+                    content
```
tsl-ps2, line 113: Do we also need something like an edited flag? How else are you generating the "Edited" labels on the FE?

tsl-ps2 (commented):
Do we want some kind of live updates for users that have the chatroom open when someone edits a message? In this case, we would need a subscription, I guess (or send updates over the already existing MessageListSubscription)?

baseapp-chats/baseapp_chats/graphql/mutations.py
```diff
@@ -390,6 +391,73 @@ def mutate_and_get_payload(
...
+                ]
+            )
+
+        content = input.get("content")
+        if len(content) < 1:
+            return ChatRoomEditMessage(
+                errors=[
+                    ErrorType(
+                        field="content",
+                        messages=[_("You need to write something")],
```
pt-tsl, line 434: Maybe change this error message to something like "You cannot edit an empty message"

baseapp-chats/baseapp_chats/graphql/mutations.py
```diff
@@ -390,6 +391,73 @@ def mutate_and_get_payload(
...
+                    )
+                ]
+            )
+
+        if len(content) > 1000:
+            return ChatRoomEditMessage(
+                errors=[
+                    ErrorType(
+                        field="content",
+                        messages=[_("You can't write more than 1000 characters")],
```
pt-tsl, line 444: I'd also change this to be something like "Message must be no longer than 1000 characters"

baseapp-chats/baseapp_chats/graphql/mutations.py
```diff
@@ -390,6 +391,73 @@ def mutate_and_get_payload(
...
+                    )
+                ]
+            )
+
+        message.content = content
+        message.save()
```
pt-tsl, line 450: Since we're only updating one field here, it makes sense to use `update_fields=["content"]`. It's a minor gain, but a gain nonetheless

pt-tsl (approved):

baseapp-chats/baseapp_chats/graphql/mutations.py
```diff
@@ -390,6 +391,73 @@ def mutate_and_get_payload(
...
+    @classmethod
+    @login_required
+    def mutate_and_get_payload(cls, root, info, **input):
+        pk = get_pk_from_relay_id(input.get("id"))
+        message = Message.objects.get(pk=pk)
+
+        profile = (
+            info.context.user.current_profile
+            if hasattr(info.context.user, "current_profile")
+            else (info.context.user.profile.pk if hasattr(info.context.user, "profile") else None)
```
tsl-ps2, line 410: I guess this should be `info.context.user.profile` (without `.pk`), right? (I don't think having a profile in the if statement and a primary key in the else statement makes much sense, but I definitely saw this at other places in the code, too, so maybe I am missing something?)

tsl-ps2 (approved):

baseapp-chats/baseapp_chats/graphql/mutations.py
Ronan-Fernandes: you are actually right the profile variable shold hold the profile object and not the PK, i think this only worked because it is falling under the first condition.

i'm also using profile.id on the permission to check. 
so I'll update this

PR #230, opened 2025-01-22
BA-1614: Multiple Profiles - Removing Member

joaosoutto:
As a user, on the BaseApp Profile,I would like to remove an existing member, In order to prevent them from doing anything in the profile.

 

Original Story:  
Acceptance Criteria 
Business Rules - Removing Member - 2

Given the user wants to remove a member, they can choose the remove option in the dropdown.

The remove option should be prominently displayed and in red font to dintinguish it from other options.

Given the user has selected to remove the member, display a confirmation dialog to make sure he intends to do this change.

Given the user has confirmed the removal of member, then remove the member from the organization and he should not be able to access the organizations profile anymore.

Copy Text

Removal Confirmation:  Are you sure you want to remove this member? This action will revoke access to the organizational profile 

Current behavior 

Expected behavior 

Code Snippet 

Design Link: https://www.figma.com/design/XRD6wSl1m8Kz6XUcAy5CLp/BaseApp---WEB?node-id=3680-68785&node-type=instance&t=4tZzXNioOJs6iIeL-0 

Approvd 
https://app.approvd.io/projects/BA/stories/35056 

DEMO: https://www.loom.com/share/c267ee0a8eb14ae787dc174abfd67c4b


vitorguima (approved):
@joaosoutto can you add a comment to this line? what do you think?

https://github.com/silverlogic/baseapp-backend/blob/546467543c21941ca33ef247ab80db0b549e6915/baseapp-profiles/baseapp_profiles/permissions.py#L51

` # Owner and admins can update roles and exclude members`

vitorguima (changes requested):
Raise package version to `0.3.9`

https://github.com/silverlogic/baseapp-backend/blob/546467543c21941ca33ef247ab80db0b549e6915/baseapp-profiles/setup.cfg#L3

release new version on pypi then update the version in the baseapp-template PR

vitorguima (approved):

baseapp-profiles/baseapp_profiles/graphql/mutations.py
```diff
@@ -156,6 +156,37 @@ def mutate_and_get_payload(cls, root, info, **input):
...
+        if not obj:
+            raise GraphQLError(_("User role not found"))
+
+        if not info.context.user.has_perm("baseapp_profiles.change_profileuserrole", obj.profile):
+            raise GraphQLError(
+                str(_("You don't have permission to perform this action")),
+                extensions={"code": "permission_required"},
+            )
+
+        id_to_return = obj.id
```
nossila, line 184: this needs to be `obj.relay_id` instead

nossila (changes requested):
Fix the return for `ProfileRemoveMember.deleted_id`

baseapp-profiles/baseapp_profiles/graphql/mutations.py
joaosoutto: @nossila it's fixed! Could you please take another look?

nossila (approved):

priscilladeroode (approved):

PR #231, opened 2025-01-24
baseapp-profiles [BA-2172] Do not overwrite profile name with user name

tsl-ps2:
**Description**

**Acceptance Criteria** 
From settings > profile - change username and save changes. The changes will be shown there. 
But if you logged in to different browser or incognito mode, the changes is not showing there.

**Video**: [https://www.loom.com/share/f60d3b6f98f04d75afc18441f7354cd8?sid=bd4f7b5c-0533-413e-9199-d598acdcee53](https://www.loom.com/share/f60d3b6f98f04d75afc18441f7354cd8?sid=bd4f7b5c-0533-413e-9199-d598acdcee53)
Bug Catcher: Selim Reza 

**Approvd** 
[https://app.approvd.io/silverlogic/BA/stories/38227](https://app.approvd.io/silverlogic/BA/stories/38227)

### DIFF
--- a/baseapp-profiles/baseapp_profiles/signals.py
+++ b/baseapp-profiles/baseapp_profiles/signals.py
@@ -1,25 +1,44 @@
 import swapper
 from django.contrib.contenttypes.models import ContentType
+from django.core.exceptions import ObjectDoesNotExist
 
 
-def update_user_profile(instance, **kwargs):
-    Profile = swapper.load_model("baseapp_profiles", "Profile")
-    target_content_type = ContentType.objects.get_for_model(instance)
+def update_user_profile(instance, created, **kwargs):
+    """
+    Call django.db.models.signals.post_save.connect(update_user_profile, sender=User)
+    to automatically create a user profile when an instance of User is saved for the first time
+    """
 
-    name = instance.get_full_name()
+    if created and not instance.profile_id:
+        Profile = swapper.load_model("baseapp_profiles", "Profile")
+        target_content_type = ContentType.objects.get_for_model(instance)
 
-    profile, created = Profile.objects.get_or_create(
-        owner=instance,
-        target_content_type=target_content_type,
-        target_object_id=instance.pk,
-        defaults=dict(
+        name = instance.get_full_name()
+
+        profile = Profile.objects.create(
             name=name,
-        ),
-    )
+            owner=instance,
+            target_content_type=target_content_type,
+            target_object_id=instance.pk,
+        )
 
-    if created:
         instance.profile = profile
         instance.save(update_fields=["profile"])
-    else:
-        profile.name = name
-        profile.save(update_fields=["name"])
+
+
+def update_user_name(instance, created, **kwargs):
+    """
+    Call django.db.models.signals.post_save.connect(update_user_name, sender=Profile)
+    to update the first_name of the user object associated to a Profile instance
+    whenever the profile name is updated.
+    """
+
+    try:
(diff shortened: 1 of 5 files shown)


tsl-ps2:
The issue with the deleted code was that each time a user was saved, the profile name on the user's standard profile was updated with the user name. So when changing the profile name in the form it was saved as expected, but when logging in on a different browser, the user was updated to keep track of the last login, and in doing so the profile name was overwritten.

I wonder whether we want "reverse" name updates, i.e. that saving a profile that is the standard profile of a user will update the user's name (there is currently no way to update user names on the front-end). The answer to that depends on the connection between users and profiles. Should every user have a unique "standard" profile (that may be used by other users, but only if it is not their standard profile)? In this case it would make sense to keep the user and standard profile name synced. But if different users can have the same "standard" profile, then such reverse name updates would not make sense (we do not want to update the names of those users sharing the profile to the same profile name).

nossila:
@tsl-ps2 ideally the user's name and profile should be synced, and eventually remove the user's name fields so we don't have to deal with that

baseapp-profiles/baseapp_profiles/signals.py
```diff
@@ -1,25 +1,43 @@
...
+    to update the first_name of the user object associated to a Profile instance
+    whenever the profile name is updated.
+    """
+
+    try:
+        user = instance.user
+    except ObjectDoesNotExist:
+        return
+
+    if user.first_name != instance.name:
```
pt-tsl, line 41: If a user wanted to change only his first name on his profile, this change would effectively duplicate the last name for the user, right? Wouldn't it be better to split the string and apply each part accordingly?

pt-tsl (changes requested):

baseapp-profiles/baseapp_profiles/signals.py
tsl-ps2: The problem is that we are not very consistent at the moment with using names. If I got this right (please correct me if I didn't), the sign up form only has one field for the name, which is saved as `first_name` (and `last_name` is always empty). In the long run, both fields should be deprecated.

But you are right, if someone still has a last name from using a previous version, then this might duplicate the name. I will set the last name to the empty string to avoid duplicating it. Is this fine?

nick-prat: I'm not sure if I agree with deleting the last name, we can have logic that depends on the last name of a user, and then also in the same vein, making the user's first name be the profile's whole name, could interfere with logic that depends on the user's first name only.  Possibly for example, we would want to list user's as `last name, first name` and sort by that, this would break that logic.  I think as a whole, I'm quite skeptical of only having a 'name' field on the profile be the single source of truth for a user's name...  I think this needs some more thought.

nick-prat: Also, if we do decide to keep this logic, this is going to update the user's name information when _any_ of their profile's names are changed, I believe what we would want to do here is get the user as `User.objects.filter(profile=instance)` and update that user if it exists.

nick-prat: Ah yes, that's correct

nick-prat (approved):

nick-prat:
Name stuff is probably out of context for this PR, we can alter this behavior later on when/if we change it.

pt-tsl (approved):

vitorguima (approved):

PR #238, opened 2025-02-03
baseapp-chats: add members to existing group

Ronan-Fernandes:
**Acceptance Criteria 
Context**
The only instance where we add members to groups is at the creation of the group, but in this story we want to enable adding members to an existing group (after creation)

**Business Rules - Add Member List**

- Adding a member is only enabled to admins
- Given I open the edit group page, I should see an "Add Member" option in the Members List section.
- Given I've clicked on the Add Member option, then the "Add Member Modal" should appear with a list of Contacts I can add to the Group
  - The list should be sorted alphabetically
  - The list should be scrollable vertically
  - Each Contact card should display their profile pic, name and username

- While on the Add Member List, I should be able to select multiple contacts to add at once.
- Make sure that the scroll bar can be selectable in the mobile view so that users don't have to manually scroll in order to get to the bottom for example.
  - We don't need to load all the contacts.
- While on the Add Member List, I should be able to use a search bar to look for a specific contact.
  - Reuse the search bar component from prior messages implementation.
- Given I have selected at least 1 Contact then enable the "confirm" button.
- Given I have confirmed the addition of the selected members then they should be added to the group and appear in the Members Section.
- Given I am on the Add Members List, when I click the cancel button, then the Add Member Modal should close and the selections should not be saved if opened again.

**Guardrails**

- In this story we will even show already existing members in the Add member list. QA and other developers should know that it might break if they select an existing member to be added again. We will fix this in another story.
- In summary we are doing the following in this story:

  - Create the Add Member button in Edit Group Page
    **Create the Add Member Modal
  - Create Add Member List with search bar and selection box
  - Logic to add the selected contacts to the group

**Design Link:** [BaseApp - WEB](https://www.figma.com/design/XRD6wSl1m8Kz6XUcAy5CLp/BaseApp---WEB?node-id=7019-410113&t=Hs9bSA9tuXgxsPGa-0)

### DIFF
--- a/baseapp-chats/baseapp_chats/graphql/mutations.py
+++ b/baseapp-chats/baseapp_chats/graphql/mutations.py
@@ -198,6 +198,7 @@ class ChatRoomCreate(RelayMutation):
 class ChatRoomUpdate(RelayMutation):
     room = graphene.Field(ChatRoomObjectType._meta.connection.Edge)
     removed_participants = graphene.List(ChatRoomParticipantObjectType)
+    added_participants = graphene.List(ChatRoomParticipantObjectType)
 
     class Input:
         room_id = graphene.ID(required=True)
@@ -232,10 +233,23 @@ class ChatRoomUpdate(RelayMutation):
             == 1
         )
         is_leaving_chatroom = profile_id in remove_participants and len(remove_participants) == 1
-
         add_participants_pks = [
-            get_pk_from_relay_id(participant) for participant in add_participants
+            get_pk_from_relay_id(participant)
+            for participant in add_participants
+            if isinstance(participant, str)
         ]
+
+        # if relay id is not valid return error
+        for participant in add_participants_pks:
+            if not participant:
+                return ChatRoomUpdate(
+                    errors=[
+                        ErrorType(
+                            field="add_participants",
+                            messages=[_("Some participants are not valid")],
+                        )
+                    ]
+                )
 
         remove_participants_pks = [
             get_pk_from_relay_id(participant) for participant in remove_participants
@@ -243,7 +257,6 @@ class ChatRoomUpdate(RelayMutation):
         participants_to_remove = ChatRoomParticipant.objects.filter(
             profile_id__in=remove_participants_pks, room=room
         )
-        removed_participants = list(participants_to_remove)
 
         title = input.get("title", None)
         image = info.context.FILES.get("image", None)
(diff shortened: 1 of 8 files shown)


baseapp-chats/baseapp_chats/graphql/mutations.py
```diff
@@ -233,6 +234,10 @@ def mutate_and_get_payload(
         )
         is_leaving_chatroom = profile_id in remove_participants and len(remove_participants) == 1
 
+        add_participants_profile = [
+            get_obj_from_relay_id(info, participant) for participant in add_participants
```
tsl-ps2, line 238: If a participant does not exist, this list may contain None, right? ...

baseapp-chats/baseapp_chats/graphql/mutations.py
```diff
@@ -320,9 +325,36 @@ def mutate_and_get_payload(
...
+        # Adding new participants
+        unique_participants_profile = list(set(add_participants_profile))
+        existing_participants = ChatRoomParticipant.objects.filter(
+            Q(room=room) & Q(profile__in=unique_participants_profile)
+        ).values_list("profile", flat=True)
+
+        new_participants = [
+            participant
+            for participant in unique_participants_profile
+            if participant.pk not in existing_participants
```
tsl-ps2, line 337: ... and here we may try to access `pk` of `None` if the participant did not exist above?

baseapp-chats/baseapp_chats/graphql/mutations.py
```diff
@@ -320,9 +325,36 @@ def mutate_and_get_payload(
...
+        ]
+        created_participants = ChatRoomParticipant.objects.bulk_create(
+            [
+                ChatRoomParticipant(
+                    profile=participant,
+                    room=room,
+                    role=(
+                        ChatRoomParticipantRoles.ADMIN
+                        if participant == profile
```
tsl-ps2, line 346: Does this make sense or was it copy and paste from `ChatRoomCreate`, where the participant who created the room automatically gets admin permissions?

tsl-ps2 (changes requested):
Would it make sense to also add a test for adding members?

baseapp-chats/baseapp_chats/graphql/mutations.py
Ronan-Fernandes: Nice catch, it was a copy paste, but in this case i think everyone has to be member.

baseapp-chats/baseapp_chats/graphql/mutations.py
Ronan-Fernandes: yeah the function was actually returning an empty string, i corrected that and made a test for it, thanks for the comment

baseapp-chats/baseapp_chats/graphql/mutations.py
Ronan-Fernandes: using only pks now to avoid errors

pt-tsl (approved):

baseapp-chats/baseapp_chats/graphql/mutations.py
```diff
@@ -234,9 +235,23 @@ def mutate_and_get_payload(
         is_leaving_chatroom = profile_id in remove_participants and len(remove_participants) == 1
 
         add_participants_pks = [
-            get_pk_from_relay_id(participant) for participant in add_participants
+            get_pk_from_relay_id(participant)
```
nick-prat, line 238: If we have to pass relay id's for this, wouldn't this make it really difficult to integrate with our API outside of relay?  (Might be a question that's outside the scope of this PR)

baseapp-chats/baseapp_chats/graphql/mutations.py
```diff
@@ -320,9 +335,35 @@ def mutate_and_get_payload(
...
+
+        created_participants = ChatRoomParticipant.objects.bulk_create(
+            [
+                ChatRoomParticipant(
+                    profile=participant,
+                    room=room,
+                    role=ChatRoomParticipantRoles.MEMBER,
+                    accepted_at=timezone.now(),
+                )
+                for participant in new_participants_profiles
```
nick-prat, line 360: you could probably just do `for participant_id in new_participants`, then `profile_id=participant_id` on line 355, and then get rid of the query on line 350.

baseapp-chats/baseapp_chats/tests/test_graphql_mutations.py
```diff
@@ -1281,3 +1287,182 @@ def test_member_user_cannot_remove_other_members(
...
+    room_id = content["data"]["chatRoomCreate"]["room"]["node"]["id"]
+    assert content["data"]["chatRoomCreate"]["room"]["node"]["participantsCount"] == 2
+
+    response = graphql_user_client(
+        UPDATE_ROOM_GRAPHQL,
+        variables={
+            "input": {
+                "profileId": django_user_client.user.profile.relay_id,
+                "roomId": room_id,
+                "addParticipants": [friend_2.relay_id + "123", friend_3.relay_id],
```
tsl-ps2, line 1320: Why does this test pass if you append "123" to `friend_2`'s relay id?

baseapp-chats/baseapp_chats/tests/test_graphql_mutations.py
```diff
@@ -1281,3 +1287,182 @@ def test_member_user_cannot_remove_other_members(
...
+        },
+    )
+    content = response.json()
+    assert (
+        content["data"]["chatRoomUpdate"]["errors"][0]["messages"][0]
+        == "Some participants are not valid"
+    )
+
+
+def test_member_user_cant_add_particioants(django_user_client, graphql_user_client, django_client):
```
tsl-ps2, line 1431: typo participants

tsl-ps2 (approved):
Left two more minor comments, but apart from that this is looking good to me.

baseapp-chats/baseapp_chats/graphql/mutations.py
Ronan-Fernandes: Yeah good point. I think we would have to adapt all other functionalities as well. since we are always using relay_id

baseapp-chats/baseapp_chats/tests/test_graphql_mutations.py
Ronan-Fernandes: No ideia this shuold not be there, it is probably a copy and paste error

nossila (approved):

PR #241, opened 2025-02-05
baseapp-chats [BA-2213]: delete messages

pt-tsl:
**Acceptance Criteria**

- Given I hover my mouse over a message, when the overlay 3 dots menu appears and I click on it, then I should see a "Delete" option in the menu. (Should already be implemented)
- Given I click the "Delete" option then, a dialog/delete message should confirm if I am sure that I want to delete the message.
  - The confirmation message changes if it's an individual chat or a group chat
- Given I click the "Delete" option to confirm, the text should be replaced with "You deleted this message" in gray font. 
  - For others in the chat (individual or in a group) the text should be replaced with "This message was deleted." in gray front. 
- A toast message should appear notifying that the message was deleted. 
  - Use the existing info icon, not the delete icon for now, nor the Undo button.  
  - Make the component as similar as the design as possible.
  - Use the new timed snack-bar component

**Guardrail**
The snackbar will not have the action button for now 

**Design Link**: https://www.figma.com/design/XRD6wSl1m8Kz6XUcAy5CLp/BaseApp---WEB?node-id=7763-127162&t=MOMbprMUFediIBZh-0 

**Note: The snackbar has been implemented in another PR**
https://github.com/silverlogic/baseapp-frontend/pull/207


baseapp-chats/baseapp_chats/graphql/mutations.py
```diff
@@ -505,6 +505,72 @@ def mutate_and_get_payload(cls, root, info, **input):
...
+            )
+
+        profile = (
+            info.context.user.current_profile
+            if hasattr(info.context.user, "current_profile")
+            else (info.context.user.profile if hasattr(info.context.user, "profile") else None)
+        )
+
+        if not info.context.user.has_perm(
+            "baseapp_chats.change_message",
```
tsl-ps2, line 547: Do we want to use the same permissions as for changing a message here? From the ACs, it does not become clear to me who should be able to delete messages. I think, it would perfectly make sense to allow an admin of a group chat to delete messages (in case they were posted by some "unwanted" participant, that the admin may remove), but not to edit messages.

baseapp-chats/baseapp_chats/graphql/object_types.py
```diff
@@ -65,6 +65,7 @@ class Meta:
             "extra_data",
             "in_reply_to",
             "is_read",
+            "deleted",
```
tsl-ps2, line 68: I guess this means participants can still query the deleted message. It might not be displayed on the frontend, but anyone knowing how to use developer tools can still see it in the response to the graphql requests. Do we care about that?

tsl-ps2: If you wanted to change this, you can just add `if root.deleted: return None` to the `resolve_content` method below.

baseapp-chats/baseapp_chats/graphql/mutations.py
```diff
@@ -505,6 +505,72 @@ def mutate_and_get_payload(cls, root, info, **input):
...
+                        field="id",
+                        messages=[_("You don't have permission to update this message")],
+                    )
+                ]
+            )
+
+        message.deleted = True
+        message.save(update_fields=["deleted"])
+
+        ChatRoomOnNewMessage.new_message(room_id=message.room.relay_id, message=message)
```
tsl-ps2, line 565: Would it make sense to rename this? (I guess this could already have been done in the edit message PR.)

tsl-ps2: Also, I think that `ChatRoomOnNewMessage.new_message` will move up the chatroom in the chatroom list one the frontend (because it triggers `ChatRoomOnRoomUpdate.new_message`). I guess we probably do not want this to happen (also for editing messages?)

baseapp-chats/baseapp_chats/graphql/object_types.py
pt-tsl: Makes sense, I'll change that

baseapp-chats/baseapp_chats/graphql/mutations.py
nossila: @pt-tsl lets change the perm to `baseapp_chats.delete_message` to make it more explicit what it is about. Its better to be very explicit in the code even though we will be doing the same checks in the permission class

baseapp-chats/baseapp_chats/graphql/mutations.py
```diff
@@ -505,6 +505,72 @@ def mutate_and_get_payload(cls, root, info, **input):
         )
 
 
+class ChatRoomDeleteMessage(RelayMutation):
+    deleted_message = graphene.Field(MessageObjectType._meta.connection.Edge)
```
nossila, line 509: There is no need to return the deleted_message, you are required to return only its relay ID to make relay in the frontend remove the message from the store. This follows better the relay/graphql pattern.

nossila (changes requested):

baseapp-chats/baseapp_chats/graphql/mutations.py
tsl-ps2, line 547: This should be "baseapp_chats.delete_message", right?

tsl-ps2 (approved):
Left one last minor comment which should take only a few seconds to fix.

Ronan-Fernandes (approved):

nossila (approved):

PR #245, opened 2025-02-07
baseapp-chats [BA-2236]: Change message subscription to not publish to sender

tsl-ps2:
**Acceptance Criteria** 
**Context**
Loom: [https://www.loom.com/share/9bcd130aa70f476fa3cebbb53ba338e0](https://www.loom.com/share/9bcd130aa70f476fa3cebbb53ba338e0)

**Business Rules**

- Given I input text into the message box, when I send the message, then the message box should become empty immediately

- Given I have sent a message, the message should be displayed in the chat history quickly.

- If a message fails to be send, just remove the message from the chat history.

  - Display a toast informing this to the user.

- Disable the send button until we know the message sending request was resolved, either successful or failed, 

**Approvd** 
[https://app.approvd.io/silverlogic/BA/stories/38918](https://app.approvd.io/silverlogic/BA/stories/38918)

### DIFF
--- a/baseapp/baseapp_chats/graphql/subscriptions.py
+++ b/baseapp/baseapp_chats/graphql/subscriptions.py
@@ -3,7 +3,7 @@ import graphene
 import swapper
 from channels.db import database_sync_to_async
 
-from baseapp_core.graphql import get_obj_from_relay_id
+from baseapp_core.graphql import get_obj_from_relay_id, get_pk_from_relay_id
 
 Profile = swapper.load_model("baseapp_profiles", "Profile")
 ChatRoom = swapper.load_model("baseapp_chats", "ChatRoom")
@@ -97,24 +97,31 @@ class ChatRoomOnMessage(channels_graphql_ws.Subscription):
 
     class Arguments:
         room_id = graphene.ID(required=True)
+        profile_id = graphene.ID(required=True)
 
     @staticmethod
-    def subscribe(root, info, room_id):
-        room = database_sync_to_async(get_obj_from_relay_id)(info, room_id)
-
+    async def subscribe(root, info, room_id, profile_id):
+        room = await database_sync_to_async(get_obj_from_relay_id)(info, room_id)
         user = info.context.channels_scope["user"]
+        profile = await database_sync_to_async(get_obj_from_relay_id)(info, profile_id)
+
         if not user.is_authenticated or not database_sync_to_async(user.has_perm)(
-            "baseapp_chats.view_chatroom", room
+            "baseapp_profiles.use_profile", profile
         ):
+            return []
+
+        if not database_sync_to_async(room.participants.filter(profile=profile).exists)():
             return []
         return [room_id]
 
     @staticmethod
-    def publish(payload, info, room_id):
+    def publish(payload, info, room_id, profile_id):
         message = payload["message"]
         user = info.context.channels_scope["user"]
 
         if not user.is_authenticated:
+            return None
+        if str(message.profile_id) == get_pk_from_relay_id(profile_id):
             return None
 
         return ChatRoomOnMessage(message=MessageObjectType._meta.connection.Edge(node=message))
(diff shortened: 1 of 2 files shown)


baseapp-chats/baseapp_chats/graphql/mutations.py
```diff
@@ -355,6 +356,22 @@ class Input:
     def mutate_and_get_payload(
         cls, root, info, room_id, content, profile_id, in_reply_to_id=None, **input
     ):
+        # Uncomment the next line if you want the sent message to appear after 5 seconds
```
tsl-ps2, line 359: I just left these comments here for testing purposes. I will remove them before merging the PR.

tsl-ps2:
Some context: I changed the new message subscription so that the sender of the message is not getting an update through the new message subscription. Before, the sender of the message received the new message both as payload to the mutation they send and through the subscription. This was no problem since relay automatically took care of "merging" the two (in this case, just ignoring the one that arrived later).

Using optimistic updates when committing the send message mutation on the frontend, these updates are only rolled back when the mutation for this payload is received, but not when the new message subscription triggers. So if the subscription reached the client before the payload, then there was a UI spike of the message showing up twice (once through the optimistic update and once through the subscription, which could not be "merged" by relay as the optimistic update has no id set) until the payload arrived (which rolled back the optimistic update and could be merged with the subscription response).  The PR fixes this behavior by not publishing the new message to its sender in the subscription.

baseapp-chats/baseapp_chats/tests/test_graphql_subscriptions.py
```diff
@@ -86,14 +86,16 @@ async def test_build_absolute_uri_on_graphql_subscription(
     # Establish & initialize WebSocket GraphQL connection.
     client = await graphql_ws_user_client(consumer_attrs={"strict_ordering": True})
 
+    print(django_user_client.user.profile)
+    print(django_user_client.user.profile.relay_id)
```
Ronan-Fernandes, line 90: remember to remove the prints

Ronan-Fernandes (approved):

nossila (approved):

pt-tsl (approved):

baseapp-chats/baseapp_chats/graphql/subscriptions.py
```diff
@@ -96,25 +96,32 @@ class ChatRoomOnMessage(channels_graphql_ws.Subscription):
...
         user = info.context.channels_scope["user"]
+        profile = await database_sync_to_async(get_obj_from_relay_id)(info, profile_id)
+
         if not user.is_authenticated or not database_sync_to_async(user.has_perm)(
-            "baseapp_chats.view_chatroom", room
+            "baseapp_profiles.use_profile", profile
         ):
             return []
+
+        if not database_sync_to_async(room.participants.filter(profile=profile).exists)():
```
nossila, line 112: please lets change this to be like before:

```suggestion
        if not database_sync_to_async(user.has_perm)(
            "baseapp_chats.view_chatroom", room
        )
```

it not only checks if the person is a member but also if he is blocked or not

nossila (changes requested):

baseapp-chats/baseapp_chats/graphql/subscriptions.py
tsl-ps2: @nossila: I am not sure whether we necessarily need to check this again since it is only for real-time updates about sent messages. When a message is sent, we check for permission (including blocks), so I think it'd be reasonable to assume that we can always synchronize messages sent on the server with the client without checking again. Also `user.has_perm` only checks whether some profile of accessible to `user` is in the chatroom, but not the given `profile`. I'll switch to 
```
if not database_sync_to_async(room.participants.filter(profile=profile).exists)() 
    or not database_sync_to_async(user.has_perm)("baseapp_chats.view_chatroom", room):
...
```

tsl-ps2: I think it would be good to rethink some of these permissions. E.g. all chat features are related to profiles, not users. So permissions like `view_chatroom` should use the `profile`, not just `user`. Also, I am currently not allowed to view a chat if there is any user blocked by me on a chat. For large chat groups, this might not make much sense; I might even lose access to all previous messages from a chat if such a user joins the group.

nossila (approved):

PR #255, opened 2025-02-17
baseapp-profiles[BA-2244]: edit/generate username for Profile

pt-tsl:
**Acceptance Criteria 
Context**
Even though the username field exists, right now we don't have any way for users to setup their username in BA. And they are important for the different social feature that we have.

**Business Rules - Automatic Username Creation**

- Given I have successfully sign up and my user is created, automatically create a username using the name. (no spaces)
- If the generated username already exists, append a number at the end, incrementing the number until a unique username is established.
- The username must be a minimum of 8 characters long. If the name is shorter than 8 characters, append random digits until it meets the requirement.
- Users should not contain special characters

**Business Rules - Edit Username**

- Parse slashes on username and profile page so that the user doesn't have to put a "/" on its user to be used in the paths.
- *Include Slashes on the URL path in the BE
- Validate username input to check for the rules set up for user creation
- Given the user inputs a username that is already in use, suggest 1 other username.

**Functional Requirements**

- The system must automatically generate a username upon successful signup.
- The system must check the uniqueness of the generated username against the existing usernames in the database

**Approvd**
https://app.approvd.io/silverlogic/BA/stories/39057

### DIFF
--- a/baseapp-profiles/baseapp_profiles/models.py
+++ b/baseapp-profiles/baseapp_profiles/models.py
@@ -1,4 +1,5 @@
-import uuid
+import random
+import string
 
 import swapper
 from baseapp_core.graphql.models import RelayModel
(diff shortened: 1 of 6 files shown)


baseapp-auth/baseapp_auth/rest_framework/register/serializers.py
```diff
@@ -26,6 +28,13 @@ def validate_referral_code(self, referral_code):
                 raise serializers.ValidationError(_("Invalid referral code."))
         return referral_code
 
+    def validate_name(self, name):
```
nossila, line 31: For me this is a bad pattern, some people do have punctuation on their names. Can we remove it?

baseapp-profiles/baseapp_profiles/models.py
```diff
@@ -109,17 +110,37 @@ def get_all_users(self):
...
+        # Remove whitespaces
+        name = name.translate(str.maketrans("", "", string.whitespace))
+
+        # If name is an email (which would only occur if the user's first and last names are empty during user registration),
+        # we'll remove the email domain and check if it's less than 8 characters. If it is, we'll add random digits to make it 8 characters.
+        # OBS: We're not checking for any other special chars since we've blocked it in the RegisterSerializer in baseapp-auth. If
+        # that changes, we should add a check here.
+        name = name.split("@")[0]
+        if len(name) < 8:
+            name = name + "".join(random.choices(string.digits, k=8 - len(name)))
```
nossila, line 132: What about we use django's [slugify](https://docs.djangoproject.com/en/5.1/ref/utils/#django.utils.text.slugify) here? Its safe and validated in production.

baseapp-profiles/baseapp_profiles/models.py
```diff
@@ -109,17 +110,37 @@ def get_all_users(self):
...
+        return f"/{name}"
 
     def create_url_path(self):
         from baseapp_pages.models import URLPath
 
         url_path = self.generate_url_path()
-        if URLPath.objects.filter(path=url_path).exists():
-            url_path = self.generate_url_path(random=True)
+        is_duplicate = URLPath.objects.filter(path=url_path).exists()
+        while is_duplicate:
```
nossila, line 141: the previous version was already recursive, I think this one can lead to multiple while loops running inside the other. not a big problem but I think a simple if is better here

nossila (changes requested):

baseapp-profiles/baseapp_profiles/models.py
pt-tsl: According to AC we're not slugifying (dashes are not valid chars for usernames)
I think it makes sense to slugify when we're dealing with URLPaths that are actually pages, rather than usernames that redirect to the user's profile.

nossila: @pt-tsl sounds good to me

Ronan-Fernandes (approved):

PR #259, opened 2025-02-25
baseapp_payments: rewrite to remove djstripe

rf2tsl:
**Customers**:

- Create a stripe endpoint handler to retrieve customer information from stripe
- Create a stripe endpoint handler to create customers in stripe
- Create a ‘Customer’ model in the DB that relates to an entity, and stores just the stripe remote customer ID, this will allow us to fetch customer data from stripe when it’s needed.
- The entity that the subscription relates to (ie. User, Organization, Profile…) should be a configurable setting defaulting to ‘Profile’
- A webhook should be implemented to receive a ‘customer.created’ which will create the related customer in our database
- A webhook should be implemented to receive a ‘customer.deleted which will delete the related customer in our database

**Subscriptions**:

- Create a stripe endpoint handler to retrieve subscription information from stripe
- Create a stripe endpoint handler to create subscriptions in stripe
- Create a ‘Subscription’ models in the DB that relates to a ‘Customer’, and stores the stripe remote subscription id, this allows us to get active subscriptions for a customer
- A webhook should be implemented to receive a ‘customer.subscription.created’ event from stripe, which will create the related ‘Subscription’ in our database
- A webhook should be implemented to receive a ‘customer.subscription.deleted event from stripe, which will delete the related ‘Subscription’ in our database

**Products**:

- Create a stripe endpoint handler to retrieve product information from stripe


baseapp/baseapp_payments/utils.py
```diff
@@ -1,138 +1,194 @@
...
+            return JsonResponse({"error": "Error"}, status=500)
+
+    @staticmethod
+    def customer_deleted(event):
+        customer_data = event["data"]["object"]
+        try:
+            Customer.objects.get(remote_customer_id=customer_data["id"]).delete()
+            return JsonResponse({"status": "success"}, status=200)
+        except Exception:
+            return JsonResponse({"status": "success"}, status=200)
```
nick-prat, line 56: Also here, not sure why we succeed on an error.

I suppose this is like this, because you're assuming that if there is an error, then the customer didn't exist in the first place, however, for that case, I think we should be explicitly checking for that here, or do `Customer.objects.filter(remote_customer_id=customer_data["id"].delete()` instead.

baseapp/baseapp_payments/utils.py
```diff
@@ -1,138 +1,194 @@
...
+        stripe.api_key = api_key
+
+    def create_customer(self, email=None):
+        try:
+            if email is None:
+                return stripe.Customer.create()
+            return stripe.Customer.create(email=email)
+        except Exception as e:
+            logger.exception(e)
+            raise ValidationError("Error creating customer in Stripe")
```
nick-prat, line 123: I'm not 100% sure if `ValidationError` is appropriate here (ditto for the other methods), since we're not actually validating anything.

nick-prat (changes requested):
Mostly looks good, requested a couple changes though.

baseapp/baseapp_payments/views.py
```diff
@@ -1,223 +1,81 @@
...
-    def start_subscription(self, request):
-        serializer = SubscribeCustomerSerializer(data=request.data, context={"request": request})
-        serializer.is_valid(raise_exception=True)
-
-        try:
+            logger.exception(e)
+            return Response({"error": "Error deleting subscription"}, status=500)
+
+    @action(detail=False, methods=["post", "get"])
+    def customer(self, request):
```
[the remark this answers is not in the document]
rf2tsl: updated to be more granular with the viewsets

baseapp/baseapp_payments/utils.py
rf2tsl: updated

baseapp/baseapp_payments/utils.py
```diff
@@ -1,138 +1,194 @@
...
+            user = get_user_model().objects.get(email=customer_data["email"])
+            if customer_entity_model != "profiles.Profile":
+                entity_model = apps.get_model(customer_entity_model)
+                entity = entity_model.objects.get(profile_id=user.id)
+            else:
+                entity_model = apps.get_model(customer_entity_model)
+                entity = entity_model.objects.get(owner=user.id)
+            Customer.objects.create(entity=entity, remote_customer_id=customer_data["id"])
+            return JsonResponse({"status": "success"}, status=200)
+        except IntegrityError:
```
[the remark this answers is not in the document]
rf2tsl: I did the success on the `IntegrityError` because of the unique constraint for creating users with the same entity and remote customer id, since the customer already exists in our db I send a 200 back to Stripe so they do not retry sending the webhook for `customer.created`. Let me know if this is bad practice or if there is a better way to handle this

nick-prat (approved):

rafaelhdr (approved):

PR #264, opened 2025-03-07
baseapp-auth: BA-2352 allow users to register using first/last name

sundayguru:
Allow for `ProfilableModel` to represent names in anyway
User should have:
first name
last name
Organization should have:
name
Profile should have:
name (editable=False)
The `Profile.name` field should be only changed via database triggers by changes on either `User` and `Organization`.
`Profile.name` should not be editable through either django admin or through the code
since profile is a more generic representation of another instances, name should be of `User`’s first name and last name or `Organization`’s marketing name, depending on which type of instance it represents.


nossila (changes requested):
We don't want to remove `Profile.name`, its supposed to be a cache from either `Organization.name` or `User.first_name + User.last_name`

`Profile.name` should never be editable directly but instead be changed by db triggers (ideally) or signals.

We need to make `Profile.name` be `editable=False` and add a `name` field to the `Organization` model.

We also need to adjust the mutations to not allow for `Profile.name` to be editable directly.

The `User` or `Organization` models will be responsible to change Profile.name based on their requirements, this way we don't need if statements on Profile to properly render the name and make it easier to customize based on your model, since any project can easily have other kinds of profiles.

baseapp-profiles/baseapp_profiles/models.py
```diff
@@ -135,6 +133,14 @@ def get_graphql_object_type(cls):
 
         return ProfileObjectType
 
+    @property
```
vitorguima, line 136: I don't think calculating this every time is a good idea since this field might be accessed through listing queries (list of comments, list of profiles, etc) where we need the profile name. This can lead to a [N+1](https://stackoverflow.com/questions/97197/what-is-the-n1-selects-problem-in-orm-object-relational-mapping) problem.  Having this as a calculated value also limits the possibility of filtering by name unless we annotate it to the queryset.

Can we please follow the PR description and use `name` as a cache field? The idea is that we update this field whenever an `organization` or `user` instance has changed.

baseapp-profiles/baseapp_profiles/models.py
```diff
@@ -135,6 +133,14 @@ def get_graphql_object_type(cls):
...
+    def name(self):
+        if hasattr(self, "user") and self.user:
+            return f"{self.user.first_name} {self.user.last_name}"
+        if hasattr(self, "organization") and self.organization:
+            return self.organization.name
```
vitorguima, line 141: this won't work since `name`  doesn't exist under `Organization`

vitorguima (changes requested):

baseapp-profiles/baseapp_profiles/models.py
sundayguru: reverted to cached field name

baseapp-profiles/baseapp_profiles/models.py
sundayguru: I introduced a name field in the org model

baseapp-profiles/baseapp_profiles/models.py
```diff
@@ -55,7 +54,7 @@ class ProfileStatus(models.IntegerChoices):
         def description(self):
             return self.label
 
-    name = models.CharField(_("name"), max_length=255, blank=True, null=True)
```
nossila, line 58: lets keep the translatable field label

nossila (changes requested):
I like where this is going. To merge this PR we need some improvements:

- [x] Add name field to `baseapp_organizations.Organization`
- [x] When updating `User.first_name`, `User.last_name` and `Organization.name` it should propagate to `Profile.name`
- [x] Add tests on both `baseapp_auth` and `baseapp_organizations` to make sure the name caching on `Profile.name` is as expected.

baseapp-chats/baseapp_chats/migrations/0011_alter_chatroomparticipant_options_and_more.py
nossila, line 1: please remove this file change since its not related to this story

nossila (approved):

vitorguima (approved):

baseapp/setup.cfg
```diff
@@ -1,6 +1,6 @@
 [metadata]
-name = baseapp_backend
-version = 0.0.23
+name = baseapp-backend
```
nossila, line 2: rename this to `baseapp_backend` please

PR #274, opened 2025-03-27
baseapp-comments: Comment Threads

tsl-ps2:
**Description**

**Acceptance Criteria** 
**Basic Comment Creation**
Since we wont have flat pages in React native and the Content feeds is not ready yet. We need a place to "host" comments in order to test them.
So we will create a comment route on the RN application to do so.

**Business Rules**

- Create a "Comment" route where we will be able to test the comment component and its functionalities

- Recreate the comment component for React Native

- Users should be able to create a comment via an input box. 

**Comment Component Behavior**

- Given a user clicks on the input box, the mobile keyboard should slide up in order for the user to type in the content.

  - The input box will slide up with the keyboard, and it should remain visible.

- Given a user presses on the Enter Key in the keyboard, it should create a line break in the content.

- Given a user has finished creating a comment, when they press the "send button", then the comment should be created.

- Given the comment has been created, then the new comment should appear at the top of the comment list.

**Reaction Component Behavior**

- Each comment should display the user’s avatar, name, and timestamp

- Users should be able to react to the comment by pression the heart icon.

- Given the user presses the Reaction Icon, the count should increase by 1

- Given the same user presses the reaction icon again after having already reacted to the comment., the count should decrease by 1.

**Guardrails**

- Threads wont be implemented for now.

- Load only the first 5 comments. Pagination will be handled later 

- Design Link: https://www.figma.com/design/z5ZdVlKEMmGIX4GVnjwfxA/BaseApp---NATIVE?node-id=942-18665&t=jbYKzqIHwPMDjdiS-0 

**Notes** 

- create a temporary model and inherit the Commentable model

- we can reuse the reply logic (e.g. `useCommentReply`) and the queries on the web 

**Approvd** 
[https://app.approvd.io/silverlogic/BA/stories/39781](https://app.approvd.io/silverlogic/BA/stories/39781)


baseapp/baseapp_comments/models.py
```diff
@@ -170,3 +170,8 @@ class CommentableModel(AbstractCommentableModel):
 
     class Meta:
         abstract = True
+
+
+# Can be used to test comments without attaching them to another model
+class CommentThread(CommentableModel, RelayModel):
```
nossila, line 176: @tsl-ps2 is the introduction to this model only for testing purposes?

PR #275, opened 2025-03-28
baseapp_reports: use ReportType model instead of ReportTypesEnum

luciobj:
✅ Social - Report Profile
Description

    As a User, on the BaseApp Profile Page,I would like to View an option to report a profile from the profile page., In order to Begin the process of reporting inappropriate behavior..

 
Acceptance Criteria 
BaseApp Profile Page

    For Desktop, tablet, and mobile

    Given I am on a user's profile page and there is a 3-dots menu, when I click on the 3-dots menu, then I should see an option labeled "Report Profile."

    When I click on "Report Profile," then a dialog should open showing the reasons to report the profile.

    The reasons should include

        Scam or Fraud

        Adult Content

        Violence, Hate or Exploitation

        Bullying or Unwanted Contact

        Other

    When I select "Adult Content" as a Reason to Report, a submenu should appear displaying these additional options

        Pornography

        Child Abuse

        Prostitution

    The submenu should allow me to select a specific sub-reason before proceeding to the text input step.

    If I select any category without available subcategories, I should directly proceed to the text input step, as they do not have submenus currently.

    Given I am in the report profile dialog, when I select a reason, then I should see a text box to elaborate on my report.

    When I enter additional details in the text box and confirm the submission, then my report should be recorded in the system.

    The system should save both the selected reason and the additional details as part of the report.

    Given I have selected a reason and provided additional details, when I confirm the submission, then I should see a confirmation dialog summarizing my report details.

    When I submit the confirmation, then I should see a success message indicating the report was submitted successfully.

    Tying the report profile flow to block profile flow will be covered in a follow-up story, but keep in mind for future implementation

Backend

    Backend should handle editable categories and subcategories for each

    Categories can have no subcategories as well

    Reports are logged on Backend

        User that submitted report 

        Profile Reported

        Date and Time of report 

Design Link: https://www.figma.com/design/XRD6wSl1m8Kz6XUcAy5CLp/BaseApp---WEB?node-id=3768-79982&node-type=section&t=7l8Nbem4bL7a0USV-0

Demo: https://www.loom.com/share/03ac73145fff4e12a513ee421d0d00b5


baseapp/baseapp_reports/graphql/object_types.py
```diff
@@ -7,27 +7,56 @@
...
+    counts = graphene.JSONString()
 
 
-def create_object_type_from_enum(name, enum):
-    fields = {}
-    for reaction_type in enum:
-        fields[reaction_type.name] = graphene.Int()
-    fields["total"] = graphene.Int()
-    return type(name, (graphene.ObjectType,), fields)
+class BaseReportTypeObjectType:
```
vitorguima, line 20: have you tried anything like:

```python
class ReportsInterface(relay.Node):
    reports_count = graphene.Field(get_reports_count_type)
```

where `get_reports_count_type` callable would handle the dynamic type creation like `create_object_type_from_enum` was handling but from the model instead?

baseapp/baseapp_reports/models.py
```diff
@@ -3,40 +3,74 @@
...
 
-    d = {"total": 0}
+def default_reports_count_full():
+    d = default_reports_count()
 
-    for report_type in ReportModel.ReportTypes:
+    ReportTypeModel = swapper.load_model("baseapp_reports", "ReportType")
+
+    for report_type in ReportTypeModel.objects.all():
         d[report_type.name] = 0
```
nossila, line 21: In the `create_object_type_from_model` you used `name.lower()` and here it is only `name`, we need to make this consistent, also  I think we should make a unique code for report_types to be used here instead of name, because name can have white spaces and any other characters. 

Can we add a test case covering this?

baseapp/baseapp_reports/graphql/object_types.py
luciobj: I did try a number of different ways to do it dynamically, but none worked as intended, some failing due to the initialization order, but most just not adding the fields as expected.

baseapp/baseapp_reports/models.py
luciobj: I changed the implementation, as the previous one was not working when building. Would you still like me to address this in any form?

nossila: @luciobj yes, there is still a bug in there, you can have a name that breaks how you access it in javascript. For example if you have `Intellectual property infringement` as a name, how do you access it in JS? You will need to make sure to always use `reports_count['Intellectual property infringement']` but if you have a safer name representation you could access it with `reports_count.intellectual_property_infringement`. Got it?

baseapp/baseapp_reports/graphql/object_types.py
```diff
@@ -7,27 +7,56 @@
...
+    class Meta(BaseReportTypeObjectType.Meta):
+        pass
 
 
 class ReportsInterface(relay.Node):
     reports_count = graphene.Field(ReportsCount)
     reports = DjangoFilterConnectionField(get_object_type_for_model(Report))
     my_reports = graphene.Field(get_object_type_for_model(Report), required=False)
 
+    def resolve_reports_count(self, info, **kwargs):
```
nossila, line 47: why do we need this resolve if we are saving it as a json field on the model itself? seems like we have a N+1 issue here that is easy to fix, seems like we could just delete this resolve method... I would also change
```reports_count = graphene.Field(ReportsCount)```
to be
```reports_count = GenericScalar()```
you can import `GenericScalar` with `from graphene.types.generic import GenericScalar`

this way on the frontend you can consume it with just `obj.reports_count.total` otherwise you would need to do something like `JSON.parse(obj.reports_count.count).total`

**Review by nossila — CHANGES_REQUESTED** [review]

File: baseapp/baseapp_reports/models.py (hunk 2 above)
Comment (luciobj) [reply baseapp/baseapp_reports/models.py]: That was always the idea, but I understand why 'name' could be confusing. Change it to 'key', hopefully it will be more intuitively that way

File: baseapp/baseapp_reports/graphql/object_types.py (hunk 3 above)
Comment (luciobj) [reply baseapp/baseapp_reports/graphql/object_types.py]: Thank you for this suggestion, it really resolved this in a great way. I have modified to it

File: baseapp/baseapp_core/graphql/object_types.py [hunk 4]
```diff
@@ -28,3 +28,7 @@ def __init_subclass_with_meta__(cls, **kwargs):
 
     def resolve_pk(self, info):
         return self.pk
+
+    @classmethod
```
vitorguima, line 32: @luciobj I guess we don't need it anymore since we are removing the django optimizer

vitorguima (changes requested):

vitorguima (approved):

nossila (approved):

nossila:
@luciobj use `squash and merge` please

PR #276, opened 2025-03-28
baseapp_payments: refactor to remove djstripe

mathieubouhelier:
Description

As a User, on the BaseApp Profile Page,I would like to subscribe to a plan through the checkout page, In order to complete my subscription payment and access premium features.

 
Acceptance Criteria 
Feature Description

Feature will be done in 2 stories - Payment Method Collection and Payment Processing

Users will reach the Subscription Checkout Page after selecting a subscription plan from the Plans Page.

The page will display:

The selected plan with pricing details.

A payment method selection area, allowing users to:

Choose from previously saved payment methods.

Add a new payment method if none exists.

A billing address input section, used for tax calculations.

A "Place Order" button, to finalize the subscription.

If the user successfully subscribes, a confirmation modal will appear.

AC - Payment Process

Given the user is on the Subscription Checkout Page, when they have saved payment methods, then the default payment method should be preselected, and clicking it should allow selecting a different method.

Given the user has selected a valid payment method, when they click "Place Order", then the system should process the payment via Stripe.

Given the user has previously saved a billing address, when they proceed to checkout, then their saved address should be pre-filled.

Given the payment is successfully processed, when the transaction is complete, then a confirmation modal should appear, displaying the order number and confirmation message.

Order Numbering should follow or use Stripe Invoice number or. ID coming from stripe and not number it on our side

Given the payment fails, when the transaction is declined, then the system should display an error message with retry options. 

Design Link: https://www.figma.com/design/XRD6wSl1m8Kz6XUcAy5CLp/BaseApp---WEB?node-id=7207-4761&t=cmCNSKhQItgRwDB7-4 

Notes 

 

Calculate tax can be covered on a follow-up story

Store everything stripe related on Stripe and we can have follow-up stories to add a caching layer to show Stripe-related information on the UI, but we should avoid storing any Stripe related info on the Admin.


baseapp/baseapp_payments/utils.py
```diff
@@ -190,3 +208,42 @@ def list_products(self, **kwargs):
...
+            logger.error(f"Stripe API error: {str(e)}")
+            return None
+
+    def list_payment_methods(self, customer_id, type="card"):
+        try:
+            payment_methods = stripe.PaymentMethod.list(
+                customer=customer_id,
+                type=type,
+            )
+            return payment_methods.data
```
nick-prat, line 229: We should probably be piping the data through our own serializer, returning exactly what we get back from stripe gives us a lot less control of our endpoint response.

baseapp/baseapp_payments/serializers.py
```diff
@@ -13,28 +13,75 @@
...
+
+    remote_subscription_id = serializers.CharField(read_only=True)
+    client_secret = serializers.CharField(read_only=True)
+    invoice_id = serializers.CharField(read_only=True)
 
     def validate(self, data):
+        customer_id = data["remote_customer_id"]
+        price_id = data["price_id"]
+
+        subscriptions = StripeService().list_subscriptions(customer_id, status="all")
```
nick-prat, line 37: I think there's a logical issue here...  I don't think it matters if a person is subscribed to two of the same price (I mean this shouldn't happen either), I think we want to make sure that the user isn't subscribed to the same **product** twice.  The way you have it here, technically a user could be subscribed to the same product, just on two different prices, if that makes sense

nossila (approved):

baseapp/setup.cfg
```diff
@@ -36,7 +36,7 @@ install_requires =
     django-ipware >= 7.0.1
     python-ipware >= 3.0.0
     requests >= 2.32.3
-
+    stripe >= 11.5.0
```
nossila, line 39: rebase please and fix this, this should be under `payments` on `extras_require` bellow

nossila (changes requested):

baseapp/baseapp_payments/serializers.py
```diff
@@ -12,29 +12,94 @@
...
+        required=False,
+    )
+
+    remote_subscription_id = serializers.CharField(read_only=True)
+    client_secret = serializers.CharField(read_only=True)
+    invoice_id = serializers.CharField(read_only=True)
+    latest_invoice = serializers.DictField(read_only=True)
 
     def validate(self, data):
+        customer_id = data["remote_customer_id"]
```
nick-prat, line 37: I might have missed it, but it looks like we're not validating that the customer is the authenticated user, can you confirm if we are, and fix if we're not?

baseapp/baseapp_payments/views.py
```diff
@@ -111,3 +124,34 @@ def retrieve(self, request, remote_customer_id=None):
...
+        try:
+            payment_methods = StripeService().get_customer_payment_methods(remote_customer_id)
+            serializer = self.serializer_class(payment_methods, many=True)
+            return Response(serializer.data, status=200)
+        except Exception as e:
+            logger.exception("Failed to retrieve payment methods: %s", e)
+            return Response({"error": "An internal error has occurred"}, status=500)
+
+    def create(self, request):
+        serializer = self.serializer_class(data=request.data)
```
nick-prat, line 147: Also here, may have missed it again, but where are we validating that the customer_id belongs to the authenticated user?

baseapp/baseapp_payments/utils.py
```diff
@@ -190,3 +219,106 @@ def list_products(self, **kwargs):
...
+        except Exception as e:
+            logger.exception(e)
+            raise Exception("Error retrieving payment methods in Stripe")
+
+    def get_customer_payment_methods(self, remote_customer_id):
+        customer = self.retrieve_customer(remote_customer_id)
+        if not customer:
+            raise Exception("Customer not found in Stripe")
```
nick-prat, line 248: We shouldn't be using the base `Exception` in this logic, we should have customer exceptions, like `CustomerNotFound`, etc...

baseapp/baseapp_payments/utils.py
```diff
@@ -190,3 +219,106 @@ def list_products(self, **kwargs):
...
+
+    def get_payment_intent(self, payment_intent_id):
+        try:
+            payment_intent = stripe.PaymentIntent.retrieve(payment_intent_id)
+            return payment_intent
+        except stripe.error.InvalidRequestError as e:
+            if "No such PaymentIntent" in str(e):
+                return None
+            logger.error(f"Failed to retrieve PaymentIntent: {str(e)}")
+            raise Exception("Error retrieving PaymentIntent in Stripe")
```
nick-prat, line 281: Also here, `PaymentMethodNotFound`

baseapp/baseapp_payments/utils.py
```diff
@@ -190,3 +219,106 @@ def list_products(self, **kwargs):
...
+    def list_payment_methods(self, customer_id, type="card"):
+        try:
+            payment_methods = stripe.PaymentMethod.list(
+                customer=customer_id,
+                type=type,
+            )
+            return payment_methods.data
+        except Exception as e:
+            logger.exception(e)
+            raise Exception("Error retrieving payment methods in Stripe")
```
nick-prat, line 243: Here too

baseapp/baseapp_payments/utils.py
```diff
@@ -190,3 +219,106 @@ def list_products(self, **kwargs):
...
+                    if pm.id == default_payment_method:
+                        pm["is_default"] = True
+                        break
+            else:
+                for pm in payment_methods:
+                    pm["is_default"] = False
+            return payment_methods
+        except Exception as e:
+            logger.exception(e)
+            raise Exception("Failed to retrieve payment methods")
```
nick-prat, line 263: Here too

baseapp/baseapp_payments/utils.py
```diff
@@ -190,3 +219,106 @@ def list_products(self, **kwargs):
...
+
+    def update_payment_method_billing_details(self, payment_method_id, billing_details):
+        try:
+            return stripe.PaymentMethod.modify(
+                payment_method_id,
+                billing_details=billing_details,
+            )
+        except Exception as e:
+            logger.exception(e)
+            raise Exception("Error updating payment method billing details in Stripe")
```
nick-prat, line 294: Here too

baseapp/baseapp_payments/utils.py
```diff
@@ -190,3 +219,106 @@ def list_products(self, **kwargs):
...
+
+        try:
+            setup_intent = stripe.SetupIntent.create(
+                customer=customer_id,
+                payment_method_types=["card"],
+            )
+            return setup_intent
+        except Exception as e:
+            logger.exception(e)
+            raise Exception("Error creating SetupIntent in Stripe")
```
nick-prat, line 306: Here too

baseapp/baseapp_payments/utils.py
```diff
@@ -190,3 +219,106 @@ def list_products(self, **kwargs):
...
+                logger.error(f"Price not found: {price_id}")
+                return None
+            logger.error(f"Invalid request for price {price_id}: {str(e)}")
+            return None
+        except stripe.error.StripeError as e:
+            logger.error(f"Stripe API error retrieving price {price_id}: {str(e)}")
+            return None
+        except Exception as e:
+            logger.exception(f"Unexpected error retrieving price {price_id}: {str(e)}")
+            raise Exception(f"Error retrieving price from Stripe: {str(e)}")
```
nick-prat, line 324: Here too

nick-prat (changes requested):
Biggest concern is in the cases where the endpoints receive a `customer_id` I don't see anywhere where we check that the `customer_id` belongs to the authenticated user.

nick-prat (approved):

PR #287, opened 2025-04-18
baseapp.content_feed: add images to posts and order_by filterset

davidbermudez-tsl:


baseapp/content_feed/graphql/queries.py
```diff
@@ -1,11 +1,21 @@
...
+    "baseapp_content_feed", "ContentPostImage", required=False, require_ready=False
+)
+ContentPostImageType = ContentPostImage.get_graphql_object_type()
+
 
 class ContentFeedQueries:
+    content_post = Node.Field(ContentPostObjectType)
     content_posts = DjangoFilterConnectionField(ContentPostObjectType)
+    content_post_image = Node.Field(ContentPostImageType)
+    content_post_images = DjangoFilterConnectionField(ContentPostImageType, max_limit=None)
```
nossila, line 21: @davidbermudez-tsl can you explain the need to expose this new fields `content_post_image` and `content_post_images`?

davidbermudez-tsl: I was starting working on the post edit. I will need to get `content_post_images` filtered by post pk to get a mappable list and put them in the Dropzone. The way it currently works is so that the content has a list of image fragments and it gets unmasked by UseFragment in a child component but since I will need to pass the list to the dropzone, the solution for this would be either adding @mask:false (not recommended by relay) or having a fragment of the images list.

davidbermudez-tsl: I've removed these changes since they won't belong in this pr anyway.

baseapp/content_feed/graphql/object_types.py
```diff
@@ -18,6 +19,11 @@
 
 
 class ContentPostImageObjectType(DjangoObjectType):
+    def resolve_image(self, info):
```
nossila, line 22: instead of this custom resolve I advise adding `image = ThumbnailField()` so the frontend can ask for the specific image sizes, I think this way the images are always the original one, right?

The profile's object type uses `ThumbnailField`

davidbermudez-tsl: We are not displaying Thumbnails here. It's the actual images, but let me see how I can optimize this images

nossila (approved):

nossila (approved):

PR #297, opened 2025-05-16
Hotfix/remove warnings

nossila:


nossila:
changed already applied

PR #307, opened 2025-06-30
feature/BA-2544-ba-be-integrate-wagtail-pages-with-base-app-modules-v…

patsluth-tsl:
Acceptance Criteria 
GraphQL Page & Block Delivery

Given the Wagtail backend with wagtail-grapple installed and configured, when a frontend queries a StandardPage, then all defined StreamField blocks should be exposed as GraphQL fragments

Given multiple StreamField block types are defined, when queried via GraphQL, then each type should be independently accessible as a fragment with type-specific fields

Given a page is queried, when the GraphQL response is returned, then it should include metadata fields like title, slug, SEO fields

Given a page is queried, when it includes children or parent pages, then the page-level context should reflect those relationships

Frontend Compatibility

Given a page exists in the Wagtail admin, when it is queried through GraphQL by the frontend, then it should render correctly using Relay without using REST endpoints

Given existing Wagtail content is present, when queried via GraphQL, then the frontend should be able to map it to React components using Relay fragments

Codebase Structure

Given the integration with wagtail-grapple, when backend services like auth and profiles are used, then they should remain fully functional and not raise integration conflicts.

Given wagtail-grapple is introduced, when existing baseapp modules are loaded, then no critical path features should break

Developer Experience

Given the GraphQL schema is implemented, when a developer inspects it, then all types and fragments should be documented using GraphQL type conventions

Given a developer is working on a frontend page, when referencing a backend-defined block, then the GraphQL schema should clearly show how that block maps to a frontend fragment

Given the GraphQL server is running, when a developer accesses GraphiQL or a similar interface, then the schema and types should be introspectable and queries testable

Proof of Success

Given the GraphQL integration is complete, when a sample page is queried and rendered on the frontend, then the end-to-end rendering should succeed with no reliance on REST endpoints

Given a page is rendered on the frontend using GraphQL, when comparing it to the REST-based rendering, then the output should be functionally equivalent 

Notes 

[SPIKE Deliverable](https://docs.google.com/document/d/1rVP4as_70u-XNVi7_8JViGUX_NkffKp0lKQm5Zqb9C4/edit?tab=t.0)

In order to connect with other baseapp packages we require a graphql implementation, this will page wagtail pages have access to:

comments

reactions

activity log (change history)

notifications

reports 

Approvd 
https://app.approvd.io/silverlogic/BA/stories/36377


baseapp_wagtail/base/graphql/object_types.py
```diff
@@ -0,0 +1,39 @@
+import graphene
+from grapple.types.interfaces import PageInterface
+
+from baseapp_comments.graphql.object_types import CommentsInterface
+from baseapp_notifications.graphql.object_types import NotificationsInterface
+from baseapp_reactions.graphql.object_types import ReactionsInterface
+from baseapp_reports.graphql.object_types import ReportsInterface
+
+
+class WagtailCommentsInterface(CommentsInterface):
```
nossila, line 10: This wrapper classes shouldn't be necessary, and there are two main issues here that made them necessary:

First `CommentsInterface`, `NotificationsInterface`, `ReactionsInterface`, and `ReportsInterface` inherits `RelayNode` and they shouldn't!

Second is that if `self.id` is `-1` is means it is [not relay complaint](https://relay.dev/docs/guides/graphql-server-specification/#object-identification), which can lead to bugs and data inconsistency on the frontend.

Can we add TODOS to the code to address this later? And/or make sure we have a story to handle this on the roadmap/epic?

baseapp_wagtail/urls.py
```diff
@@ -16,4 +17,5 @@
     # the system won't provide a final path for the pages and we need it to generate the headless
     # pages.
     path("", include(wagtail_urls)),
+    path("cms/", include(grapple_urls)),
```
nossila, line 20: Why is this needed? Seems like `grapple.urls` is just adding the `/graphql` again, but now under `/cms/graphql`. Does it break without it?

nossila (changes requested):

baseapp_wagtail/base/graphql/object_types.py
patsluth-tsl: Done

baseapp_wagtail/urls.py
patsluth-tsl: It seems there is one other url which could possibly be added by grapple.
`graphiql/`
I think it is fine if we remove this, but also fine if we keep it. It just means we can access the graphql interface from multiple urls. 

I will leave up to @Hercilio1 who is taking over this PR.

nossila (approved):

PR #311, opened 2025-08-06
Feature/ba 2640 sync payments with fe after refactor

lbjunq:
Previous refactor was not quite synched with FE due to the epic not being updated with compatible versions bewteen FE package and template. Now that they are all aligned with current master, this changes were necessary so all flow works as intended.

### DIFF
--- a/baseapp_payments/views.py
+++ b/baseapp_payments/views.py
@@ -9,6 +9,7 @@ from rest_framework.response import Response
 
 from baseapp_core.rest_framework.decorators import action
 from baseapp_core.rest_framework.mixins import DestroyModelMixin
+from baseapp_core.graphql import get_pk_from_relay_id
 
 from .models import Subscription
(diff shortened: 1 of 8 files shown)


priscilladeroode (approved):

pablogadhi (approved):

mathieu-tsl (approved):

baseapp_payments/views.py
```diff
@@ -166,15 +181,13 @@ def payment_methods(self, request, pk=None, payment_method_id=None, *args, **kwa
 
 
 class StripePaymentMethodViewset(viewsets.GenericViewSet):
-    permission_classes = [IsAuthenticated, DRFCustomerPermissions]
```
nossila, line 169: you removed permission checking here, are you sure about this?

lbjunq: I think so, since this Viewset is only used by another one, on a action that already has the permission classes.

PR #317, opened 2025-08-21
feature/BA-2492-improvements-to-products-viewset

lbjunq:
✅ Stripe Checkout - Subscription Plans Page
Description

    As a User, on the BaseApp Profile Page,I would like to view and manage available subscription plans on a responsive page tailored for different devices, In order to select or manage a plan that meets my needs effectively.

 
Acceptance Criteria 
Frontend

    Responsive Layouts

        Desktop (Horizontal layout):

            Display plans in a row, account for multiple rows if more than 3 plans exist.

            Ensure proper alignment and consistent card widths regardless of the number of plans.

        Mobile & Tablet (Vertical layout):

            Stack plans vertically to ensure usability and readability on smaller screens.

CoP comments

     

        Its ok to reuse Stripe Web Elements. 

        Check if its feasible to replace the BaseApp Components with the Stripe Payment Elements

            Check so that we can customize the individual component so that the end result is as similar as possible to BaseApp components in the designs: https://docs.stripe.com/elements/appearance-api#rules

        We have to think about what customizations are we going to be enabled so that components can be re used in other projects.

            Probably a good idea to talk to CoP when thinking of this.

Plan Cards

    Reusable plan card component:

        Title, pricing, description, and features dynamically populated via backend.

Call-to-Action Buttons

    Highlight the “Subscribe” button for available plans.

        Use of the button to redirect user to Checkout Page will be covered in a follow-up story 

Design Link: https://www.figma.com/design/XRD6wSl1m8Kz6XUcAy5CLp/BaseApp---WEB?node-id=7170-121730&t=EqObBvkTBQ7zD0Fr-4 

Approvd 
https://app.approvd.io/projects/BA/stories/38018 

Demo: In progress...


baseapp_payments/serializers.py
```diff
@@ -20,6 +20,73 @@
...
+    currency = serializers.CharField()
+    unit_amount = serializers.IntegerField()
+    recurring = serializers.DictField()
+
+
+class StripeProductSerializer(serializers.Serializer):
+    id = serializers.CharField()
+    name = serializers.CharField()
+    description = serializers.CharField(allow_null=True, required=False)
+    active = serializers.BooleanField()
```
mathieubouhelier, line 34: does it make sense to add a default value?

mathieubouhelier (approved):

baseapp_payments/serializers.py
lbjunq: Not really, as this always comes from Stripe, and the status is a kwarg on the list.

baseapp_payments/utils.py
```diff
@@ -450,3 +455,479 @@ def get_customer_invoices(self, customer_id):
         except Exception as e:
             logger.exception(e)
             raise InvoiceNotFound("Error retrieving invoices in Stripe")
+
+# instance = {
```
Hercilio1, line 459: Do we wanna keep this commented json here?

Hercilio1 (changes requested):

baseapp_payments/utils.py
lbjunq: Removed

Hercilio1 (approved):

PR #320, opened 2025-08-26
[BE - package] User Deletion - Remove User Info from Messages

mathieubouhelier:
Acceptance Criteria 
As a user in the Messaging Feature, I want messages from deleted users to preserve their content but not show their name, avatar, or profile link, so that the conversation history remains readable without exposing data from deleted accounts.

Acceptance Criteria  

When a user deletes their account, the system should anonymize their presence in all chat messages they participated in.  

The user's name and avatar should be removed from all past messages.  

Replace the avatar with a generic "Deleted User" icon.  

Display a label such as “Deleted User” instead of their name.  

The link to the deleted user's profile should be removed from all messages.  

Messages sent by deleted users should remain visible to other participants in the chat.  

Group and direct chats should continue to display messages from deleted users with anonymized attribution.  

The message layout and order should remain unchanged after account deletion.


lbjunq (approved):

joaosoutto (approved):

baseapp_auth/anonymization.py
```diff
@@ -3,6 +3,22 @@
 
 
 def anonymize_activitylog(self, *args, **kwargs):
+    """
```
vitorguima, line 6: @mathieubouhelier Do you know which metadata would be useful for audit purposes? Maybe it's worth mentioning those under this comment block.

baseapp_auth/rest_framework/users/tasks.py
```diff
@@ -31,10 +31,9 @@ def anonymize_and_delete_user_task(user_id):
                 room = ChatRoom.objects.get(id=room_id)
                 room.participants_count = ChatRoomParticipant.objects.filter(room=room).count()
                 room.save(update_fields=["participants_count"])
-
         user.delete()
         send_anonymize_user_success_email(user_email)
     except Exception as e:
         logging.exception(e)
-        send_anonymize_user_error_email(user.id)
+        send_anonymize_user_error_email(user_email)
```
vitorguima, line 38: @mathieubouhelier for the `anonymize_and_delete_user_task` I think we should explicitly control these transactions using `with transaction.atomic()`  for the whole `try` block. 

This way we make sure that every transaction was successfully completed. If any of it fails then we rollback every  changes in the flight.

doc reference: https://docs.djangoproject.com/en/5.2/topics/db/transactions/#controlling-transactions-explicitly

baseapp_auth/tests/integration/test_users.py
```diff
@@ -215,6 +224,104 @@ def test_admin_can_only_deactivate_account(self, user_client):
...
+            mock_apply_async.assert_called_once_with(
+                args=[user_id], eta=mock_apply_async.call_args[1]["eta"]
+            )
+
+    def test_anonymize_and_delete_user_task_sends_success_email(self, outbox):
+        user = UserFactory()
+        user_id = user.id
+
+        anonymize_and_delete_user_task(user_id)
+        assert len(outbox) == 1
```
vitorguima, line 246: @mathieubouhelier guess we should cover both scenarios where it should and shouldn't be triggered:

if `config.SEND_USER_ANONYMIZE_EMAIL_TO_SUPERUSERS = True`  and if `config.SEND_USER_ANONYMIZE_EMAIL_TO_SUPERUSERS = False`

baseapp_chats/graphql/object_types.py
```diff
@@ -100,7 +100,7 @@ def get_profile_pk(info, profile_id=None):
     @staticmethod
     def get_replaced_profile_name(profile, profile_pk, replacement_text):
         if not profile:
-            return None
+            return "Introuvable Profile"
```
vitorguima, line 103: @mathieubouhelier any reasons why we are returning this string?

mathieu-tsl: This was requested by the COPE frontend team — the idea is to return a string instead of None, so it can be rendered directly. (@anicioalexandre) 

@vitorguima Do you see any problem?

nossila: @mathieu-tsl we need to have it in english and wrapped on `gettext_lazy` to make this translated, this is a requirement for any user readable text/word

baseapp_auth/rest_framework/users/tasks.py
mathieu-tsl: Done ty

baseapp_auth/anonymization.py
mathieu-tsl: Upaded the comments ty @vitorguima

PR #335, opened 2025-10-07
Update: Adding default value to API Key env vars

Hercilio1:
Addressing critical issue that is blocking the staging environment from running


Janekk (approved):

baseapp_api_key/managers.py
patsluth-tsl, line 1: I think a ImproperlyConfigured exception would be more appropriate

vitorguima (approved):

PR #418, opened 2026-04-27
feat: add mentions app

anicioalexandre:
## Summary by CodeRabbit

* **New Features**
  * Users can mention profiles when creating or editing posts, chat messages, and comments; mentioned profiles are shown and tracked per item.
  * Mentions support listing, counts, and a boolean checker on content; self-mentions are excluded and malformed IDs are ignored.
  * New Mentions app and models with signals to notify changes.

* **Tests**
  * Comprehensive end-to-end and unit tests covering create/edit flows, service logic, GraphQL resolvers, and signal behavior.


baseapp_chats/tests/test_mentions.py
```diff
@@ -0,0 +1,231 @@
...
+    c = ProfileFactory()
+    message.mentioned_profiles.set([a, b])
+
+    graphql_user_client(
+        EDIT_MESSAGE_WITH_MENTIONS_GRAPHQL,
+        variables={
+            "input": {
+                "id": message.relay_id,
+                "content": "edited",
+                "mentionedProfileIds": [c.relay_id],
```
Ronan-Fernandes, line 186: It might be out of scope for this story, but should we allow profiles that are not group participants to be mentioned?

Ronan-Fernandes (approved):

baseapp_comments/graphql/object_types.py
```diff
@@ -163,6 +163,7 @@ class Meta:
             "in_reply_to",
             "language",
             "status",
+            "mentioned_profiles",
```
vitorguima: @anicioalexandre can we have some tests for understanding how optimized this field is? My concern is if it creates unnecessary queries and some overload, specially because this might be an extensive list... I also think we could have this field under a specific interface with an optimized resolver (if needed) so we can just reuse that on other apps and enable it only and if the mentions app is installed.

baseapp/content_feed/graphql/mutations.py
```diff
@@ -64,6 +67,13 @@ def mutate_and_get_payload(cls, root, info, **input):
                         )
                     )
 
+                if mentioned_profile_ids:
```
vitorguima: this creates a direct dependency between `content_post` and `mentions`. Same thing happens for the `chats`. I wonder if we should just keep some helpers inside the package and actually implement this at the application level instead.  @nossila @ho-tsl any thoughts?

baseapp_chats/tests/test_mentions.py
anicioalexandre: that's a good point, I think we can leave this for future follow ups

baseapp/content_feed/graphql/object_types.py
```diff
@@ -19,6 +20,17 @@
 )
 
 
+content_post_interfaces = (
+    RelayNode,
+    ReactionsInterface,
+)
+
+if apps.is_installed("baseapp_mentions"):
```
anicioalexandre, line 28: added this safeguard here since mention is not yet a thing on baseapp, we still need to do more work on it on both frontend and backend

Hercilio1 (approved):

baseapp_mentions/graphql/interfaces.py
```diff
@@ -0,0 +1,50 @@
...
+    No fields are added to the consuming model — the resolvers query the
+    Mention through-table keyed off the consuming object's DocumentId.
+    """
+
+    mentioned_profiles = DjangoFilterConnectionField(get_object_type_for_model(Profile))
+    mentions_count = graphene.Int()
+    is_mentioning_profile = graphene.Boolean(profile_id=graphene.ID(required=True))
+
+    def resolve_mentioned_profiles(self, info, **kwargs):
+        doc = DocumentId.get_or_create_for_object(self)
```
nossila, line 34: @Hercilio1 I think on cases like this we should use `self.public_id` no? So it follows the optimization patterns in place as well and avoids extra queries?

baseapp_mentions/models.py
```diff
@@ -0,0 +1,48 @@
...
+
+class AbstractBaseMention(TimeStampedModel, RelayModel):
+    """A profile mentioned inside a target document.
+
+    `target` is a `DocumentId` so any model registered with the DocumentId
+    registry can be a mention target without adding fields or migrations of
+    its own — mirrors the `baseapp_follows.Follow` pattern.
+    """
+
+    profile = models.ForeignKey(
```
nossila, line 18: @anicioalexandre this was who mentioned, right? Usually we also look into saving the current user for audit reasons since a Profile can be managed by multiple Users. But I think for the current cases we don't need because mentions are either in a Chat Message or CommentItem, and then both already saves the current user and active profile, making it redundant to save the profile here as well.

baseapp_mentions/graphql/interfaces.py
```diff
@@ -0,0 +1,50 @@
...
+        profile_pks = Mention.objects.filter(target=doc).values_list("profile_id", flat=True)
+        return gql_optimizer.query(Profile.objects.filter(pk__in=profile_pks), info)
+
+    def resolve_mentions_count(self, info):
```
nossila, line 38: @anicioalexandre counts are very costly on this type of features, so it is nice to cache them. We are making an intermediate table with this information, and the counts can be updated using triggers to decrease latency and improve consistency. There are examples here: https://github.com/silverlogic/baseapp-backend/blob/master/baseapp_chats/triggers.py

PR #421, opened 2026-05-06
Refactor baseapp_reports to use plugin architecture - NOT READY FOR REVIEW

tsl-ps2:
### User Story
The baseapp_ratings and baseapp_reports apps need to be aligned with the BaseApp plugin architecture and updated to remove remaining coupling patterns. While these apps are relatively self-contained, they still rely on patterns such as direct Profile references or model inheritance that limit extensibility and reuse.

This effort ensures both apps follow the plugin-driven architecture, remove unnecessary coupling, and expose functionality through standardized mechanisms such as registries and shared services where appropriate.

### Acceptance Criteria
This story is done when:

Concrete models in baseapp_ratings and baseapp_reports are reviewed and removed if they should be extendable, so that flexibility is preserved for project-level customization.

Migration folders are removed from apps that no longer contain concrete models, so that unnecessary database artifacts are eliminated.

All coupling between baseapp_ratings and baseapp_profiles (e.g., Profile FK via swapper) is removed or replaced with a decoupled reference (such as DocumentId or project-owned models), so that the ratings system is independent.

baseapp_reports remains free of coupling to other baseapp apps, and any future dependency on Profile or external models is avoided or replaced with a decoupled pattern, so that the app stays modular.

Both apps are registered as plugins and contribute their GraphQL operations and settings via the registry, so that configuration is centralized and dynamic.

Any URLs exposed by these apps are contributed via plugin callbacks, so that routing is managed through the registry.

GraphQL shared interfaces are resolved by name only, so that no direct cross-package imports are required.

If ratings expose reusable logic (e.g., average rating or counts), a shared service is introduced and consumed via the shared service registry, so that other apps can access this functionality without direct imports.

If reports functionality needs to be consumed externally (e.g., report counts or reportable checks), a shared service is introduced (e.g., get_report_count(document_id)), so that consumers like comments use registry-based access instead of inheritance patterns.

Any existing inheritance-based coupling (e.g., ReportableModel) is removed or replaced with service-based or DocumentId-based approaches, so that apps interact through decoupled contracts.

All plugin registrations and optional shared services are initialized in AppConfig.ready(), so that the system bootstraps consistently.

The final structure contains no direct cross-app dependencies and adheres to plugin architecture principles, so that both apps are reusable and maintainable.

### Approvd
https://app.approvd.io/silverlogic/BA/stories/46317

### Notes
It was decided to also cover baseapp_reactions in this story


baseapp_profiles/plugin.py
```diff
@@ -37,7 +37,6 @@ def get_settings(self) -> PackageSettings:
             optional_packages=[
                 "baseapp_blocks",
                 "baseapp_follows",
-                "baseapp_reports",
```
Hercilio1, line 40: Here I think it's okay to leave, since the profile object type still uses its interface

baseapp_reports/management/commands/update_reports_count.py
```diff
@@ -1,12 +1,35 @@
-from apps.reactions.models import ReportableModel, update_reports_count
-from django.apps import apps
+import swapper
+from django.contrib.contenttypes.models import ContentType
 from django.core.management.base import BaseCommand
 
 
 class Command(BaseCommand):
```
Hercilio1, line 7: Should we add a few tests on this command, since we are changing it, I think it's worthy doing it

baseapp_reports/migration_helpers/convert_legacy_reports_count_to_metadata_helper.py
```diff
@@ -0,0 +1,93 @@
...
+    apps,
+    schema_editor,
+    *,
+    source_app_label: str,
+    source_model_name: str,
+    metadata_app_label: str = "baseapp_reports",
+    metadata_model_name: str = "ReportableMetadata",
+    reports_count_field: str = "reports_count",
+):
+    SourceModel = apps.get_model(source_app_label, source_model_name)
```
Hercilio1, line 48: In baseapp_follows/migration_helpers/convert_follow_profile_fks_into_document_id_helper.py I'm using a thing called schema.alias (or something like that) to make use I'm running the migration in the right database. Can you take a look and try to apply the same logic here and in the other helpers too?

baseapp_reports/tests/test_graphql_queries_object_reports.py
```diff
@@ -0,0 +1,104 @@
...
+
+
+@override_config(ENABLE_PUBLIC_ID_LOGIC=True)
+def test_anon_reports_count_is_flat_regardless_of_report_volume(
+    django_user_client, graphql_client_with_queries
+):
+    """``reportsCount`` should be a flat query path: regardless of how many ``Report``
+    rows point at the target, the GraphQL query should make the same number of DB
+    round-trips. This locks in the ``ReportableMetadataService.annotate_queryset``
+    optimisation."""
```
Hercilio1: I don't know why claude is using this double backtick (``) in the docstringss. But I replace the ones in my PR by simple backticks.

testproject/reports/models.py
```diff
@@ -1,8 +1,12 @@
...
+    AbstractReportableMetadata,
+    AbstractReportType,
+)
 
 
-class ReportType(AbstractBaseReportType):
-    class Meta(AbstractBaseReportType.Meta):
+class ReportType(AbstractReportType):
+    class Meta(AbstractReportType.Meta):
         db_table = "baseapp_reports_reporttype"
```
Hercilio1: remove this db_table here. In the template u can if there we use it

testproject/reports/models.py
```diff
@@ -12,12 +16,17 @@ def get_graphql_object_type(cls):
...
-    class Meta(AbstractBaseReport.Meta):
+class Report(AbstractReport):
+    class Meta(AbstractReport.Meta):
         db_table = "baseapp_reports_report"
 
     @classmethod
     def get_graphql_object_type(cls):
         from baseapp_reports.graphql.object_types import ReportObjectType
 
         return ReportObjectType
```
Hercilio1: define this `get_graphql_object_type` in the abstraction and remove from here as well. In the template u can leave, since there is works as a placeholder for changing.

baseapp_reports/models.py
Hercilio1: Lets change this GFK to connect direct in the DocumentId. This might be a big change.

Hercilio1 (changes requested):

PR #423, opened 2026-05-08
BA-3247 Refactor Baseapp Ratings and Reports to Plugin Architecture

tsl-ps2:
User Story
The baseapp_ratings and baseapp_reports apps need to be aligned with the BaseApp plugin architecture and updated to remove remaining coupling patterns. While these apps are relatively self-contained, they still rely on patterns such as direct Profile references or model inheritance that limit extensibility and reuse.

This effort ensures both apps follow the plugin-driven architecture, remove unnecessary coupling, and expose functionality through standardized mechanisms such as registries and shared services where appropriate.

Acceptance Criteria
This story is done when:

Concrete models in baseapp_ratings and baseapp_reports are reviewed and removed if they should be extendable, so that flexibility is preserved for project-level customization.

Migration folders are removed from apps that no longer contain concrete models, so that unnecessary database artifacts are eliminated.

All coupling between baseapp_ratings and baseapp_profiles (e.g., Profile FK via swapper) is removed or replaced with a decoupled reference (such as DocumentId or project-owned models), so that the ratings system is independent.

baseapp_reports remains free of coupling to other baseapp apps, and any future dependency on Profile or external models is avoided or replaced with a decoupled pattern, so that the app stays modular.

Both apps are registered as plugins and contribute their GraphQL operations and settings via the registry, so that configuration is centralized and dynamic.

Any URLs exposed by these apps are contributed via plugin callbacks, so that routing is managed through the registry.

GraphQL shared interfaces are resolved by name only, so that no direct cross-package imports are required.

If ratings expose reusable logic (e.g., average rating or counts), a shared service is introduced and consumed via the shared service registry, so that other apps can access this functionality without direct imports.

If reports functionality needs to be consumed externally (e.g., report counts or reportable checks), a shared service is introduced (e.g., get_report_count(document_id)), so that consumers like comments use registry-based access instead of inheritance patterns.

Any existing inheritance-based coupling (e.g., ReportableModel) is removed or replaced with service-based or DocumentId-based approaches, so that apps interact through decoupled contracts.

All plugin registrations and optional shared services are initialized in AppConfig.ready(), so that the system bootstraps consistently.

The final structure contains no direct cross-app dependencies and adheres to plugin architecture principles, so that both apps are reusable and maintainable.

Approvd
https://app.approvd.io/silverlogic/BA/stories/46317

Notes
It was decided to also cover baseapp_reactions in this story

## Release Notes

* **Architecture Improvements**
  * Optimized data model architecture for ratings, reactions, and reports—reducing database queries through metadata services and efficient queryset annotations.
  * Migrated from generic foreign key patterns to document-based targeting for improved data consistency.

* **Plugin Integration**
  * Enhanced plugin-based GraphQL schema contributions for ratings, reactions, and reports interfaces.
  * Enabled modular plugin registration for authentication backends and shared services.


baseapp_comments/graphql/filters.py
```diff
@@ -12,8 +14,14 @@ class CommentFilter(django_filters.FilterSet):
...
             ("created", "created"),
             ("is_pinned", "is_pinned"),
-            ("reactions_count__total", "reactions_count_total"),
-            # replies_count_total is annotated in BaseCommentObjectType.pre_optimization_hook
+            # `reactions_count_total` is annotated in `BaseCommentObjectType.pre_optimization_hook`,
+            # if `baseapp_reactions` is installed.
+            *apply_if_installed(
+                "baseapp_reactions",
+                [("reactions_count_total", "reactions_count_total")],
+                [],
```
Hercilio1: the apply_if_installed method infers basic fallback values. So you can remove this third param with `[]`.

baseapp_ratings/tests/test_graphql_queries_object_ratings.py
```diff
@@ -0,0 +1,93 @@
...
+            }
+        }
+    }
+"""
+
+
+@override_config(ENABLE_PUBLIC_ID_LOGIC=True)
+def test_anon_ratings_count_is_flat_regardless_of_rate_volume(
+    django_user_client, graphql_client_with_queries
+):
```
Hercilio1, line 45: Wanna add a new test that explicitly checks the actual number of queries (e.g. `assert queries.count == 4`)? That way if bad affects the base query of it, we would be awere.

baseapp_ratings/models.py
```diff
@@ -30,27 +31,29 @@ class Meta:
...
-    target = GenericForeignKey("target_content_type", "target_object_id")
     value = models.IntegerField(default=0, blank=False, null=False)
 
     class Meta:
         abstract = True
+        swappable = swapper.swappable_setting("baseapp_ratings", "Rate")
+        unique_together = [["user", "target_document"]]
         indexes = [
-            models.Index(fields=["target_content_type", "target_object_id"]),
+            models.Index(fields=["target_document"]),
```
Hercilio1: Now that the GFK became a simple FK, we don't need to create this index manually anymore.

baseapp_reactions/tests/test_graphql_queries_object_reactions.py
```diff
@@ -0,0 +1,93 @@
+"""
+Query-count tests for the `ReactionsInterface`.
```
Hercilio1, line 2: Same review as we have for ratings. Lets add at least one extra test checking the queries count asserting it against a static value.

baseapp_reports/tests/test_graphql_queries_object_reports.py
```diff
@@ -0,0 +1,104 @@
+"""
```
Hercilio1, line 1: Same review as we have for ratings. Lets add at least one extra test checking the queries count asserting it against a static value.

Hercilio1 (changes requested):

Hercilio1 (approved):

baseapp_ratings/graphql/object_types.py
```diff
@@ -8,11 +8,15 @@
...
+from baseapp_core.plugins import apply_if_installed, shared_services
 
 RateModel = swapper.load_model("baseapp_ratings", "Rate")
 
 
+def _service():
+    return shared_services.get("ratable_metadata")
+
+
 class RatingsInterface(RelayNode):
```
nossila, line 20: @Hercilio1 @tsl-ps2 It would be interesting to change it to inherit from `graphene.Interface` instead, some of our interfaces are inheriting from `RelayNode` but that was a mistake, and since we are making a big refactor that might need some frontend changes, I think this is a good opportunity to fix this as well.

nossila (approved):

PR #442, opened 2026-07-12
feat(files): baseapp.files package — attach files to any object (multipart upload, plugin arch)

nossila:
## Summary

Adds the **`baseapp.files`** package — a plugin-architecture file feature that attaches files to any `DocumentId`-enabled object via a shared `FilesInterface`, with S3-style multipart upload for large files. Built on the same shared-service/annotation pattern as `baseapp_reactions`/`baseapp_comments`.

## What's included

**Core package (`baseapp/files/`)**
- Abstract, swappable models `AbstractFile` / `AbstractFileTarget` (`DocumentId`-keyed, counts maintained by pgtriggers). Concrete models live in consumers.
- Multipart upload (initiate → presigned part URLs → complete/abort) with S3 and local-dev storage handlers. **Multipart-only — resumable uploads are intentionally out of scope.**
- GraphQL: `FilesInterface`, `FileObjectType`, `FileAttachToTarget` / `FileDelete` mutations, `myFiles` / `file` queries.
- REST: upload initiate/complete/abort + local token-authenticated part upload.
- `ContentPost` now implements `CommentsInterface` so posts can carry comments (which in turn carry files).

**Performance (mirrors the reactions pattern)**
- `AbstractFileTarget.annotate_queryset` + a `files_metadata` shared service; `FilesInterface` resolvers read annotations with a fallback (no per-row `get_or_create`/`DocumentId` lookups).
- Comments' `pre_optimization_hook` applies the files annotations, so listing comments with files is flat.
- Query-count tests lock this in (`filesCount` invariant to file volume; comments-with-files listing invariant to list size; single-SELECT annotation test).

**Security / correctness**
- `FileAttachToTarget` authorizes the target (rejects non-`FileableModel` targets; delegates "may attach here" to an overridable `add_file` permission) and requires each id to resolve to a `File`.
- `myFiles` scoped to the authenticated user (was world-readable) with deterministic ordering.
- `attach_files_from_relay_ids` no longer treats a NULL creator as an implicit grant.

**Quality**
- ≥ 85% new-code coverage. SonarCloud branch quality gate is **green**: 0 open new-code issues, all ratings A, both security hotspots reviewed (CSRF-exempt token endpoint + MD5 content ETag — both confirmed safe).
- `AbstractFile` string fields use `blank=True, default=""` (no `null=True`); consumer migrations include a NULL→`""` data step for safety on populated databases.

## Test plan
- `docker compose run --rm web pytest baseapp/files/` — 118 passed (fresh DB, exercises migrations).
- Comments + content_feed suites green; `makemigrations --check` clean; black/isort/flake8 clean.
- CI ("Lint & Tests & SonarCloud") green on Python 3.11 & 3.12.

## Notes for consumers
Downstream projects bumping this library regenerate their concrete `files` migration; if their table has existing NULL string values, include the same NULL→`""` data step before the `AlterField` (see `testproject/files/migrations/0002`).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **New Features**
  * Added file listing, attachment, deletion, and re-parenting through GraphQL and REST, including file metadata on supported content.
  * Added direct multipart uploads with S3 presigned parts or local storage fallback, abort support, and automatic cleanup.
  * Added multipart/form-data support for GraphQL requests and REST Current-Profile header support.
* **Documentation**
  * Added guides for S3 multipart uploads and the complete upload flow.
* **Bug Fixes**
  * Standardized image and thumbnail GraphQL fields to return direct URL or scalar values.
* **Tests**
  * Expanded coverage for uploads, permissions, cleanup, and file-related GraphQL behavior.


nossila:
Second review round addressed in `290fbaf`:

**Security / correctness**
- **Storage backend never detected S3** — `type(default_storage)` returns the `LazyObject` wrapper, so `"S3Boto3Storage"` never matched and `get_upload_handler()` always returned the local handler. Now uses `default_storage.__class__` (which `LazyObject` proxies to the real backend). Matches the existing factory tests.
- **Residual CodeQL "information exposure"** on the `except ValueError` branches: those now log server-side and return a generic message (the specific reason is in the logs). Completion-error tests updated accordingly.
- **Local `complete_upload` could silently truncate** — a missing part file now raises instead of producing a partial file.
- **`?status=` empty value bypassed the filter** — now uses an explicit presence/empty check, so it can't leak files in non-completed states.

**Stability / hygiene**
- Cleanup tasks load the swappable `File` model lazily (avoids `AppRegistryNotReady` when the module is imported during Celery startup); the swallowed storage-abort failure in `abort_multipart_upload` is now logged with a traceback.
- `FilesInterface.resolve_files` orders by `(-created, -pk)` for stable cursor pagination, consistent with `my_files`.
- MD5 ETag marked `usedforsecurity=False` (content checksum, not a security primitive).
- DRF `ValidationError` messages wrapped in arrays / `non_field_errors`; `FilesPermissionsBackend.has_perm` is now typed and documented.
- Fixed the README multipart JS example to the single `parent_id` argument.

**Deliberately deferred** (noted, not done here):
- `help_text` on the three `CharField`s — Django tracks `help_text` in migration state, so adding it forces an `AlterField` migration across every consumer for a cosmetic attribute; not worth destabilizing the data-safe `0002` migrations in this PR.
- Moving the S3 network calls outside the `@transaction.atomic`/`select_for_update` block (the two "Heavy lift" performance notes) — a real improvement but a larger change to transaction boundaries; better as a focused follow-up than folded into this PR.
- `baseapp_profiles` `CurrentProfileMixin` missing-profile guard — outside this package's scope (pre-existing profiles code).

Files suite green (120 passed), migrations clean.

baseapp_profiles/rest_framework/mixins.py
```diff
@@ -0,0 +1,50 @@
...
+        """
+        current_profile_header = request.headers.get("Current-Profile")
+        request.user.current_profile = None
+
+        if not request.user.is_authenticated:
+            return
+
+        if not current_profile_header:
+            # No header provided, use the user's default profile
+            request.user.current_profile = request.user.profile
```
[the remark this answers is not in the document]
nossila: Out of scope: pre-existing code in a different package, not the files feature

baseapp/files/rest_framework/uploads/presigned_views.py
```diff
@@ -0,0 +1,170 @@
...
+        except signing.SignatureExpired:
+            return Response(
+                {"error": _("Token has expired")},
+                status=status.HTTP_401_UNAUTHORIZED,
+            )
+        except signing.BadSignature:
+            return Response(
+                {"error": _("Invalid token signature")},
+                status=status.HTTP_401_UNAUTHORIZED,
+            )
```
Hercilio1: We could isolate the token logic in a reusable class or utils file. What do you think?

baseapp/files/base.py
```diff
@@ -0,0 +1,31 @@
...
+    def get_file_target(self):
+        return get_or_create_file_target(self)
+
+    @property
+    def files_count(self) -> dict:
+        return self.get_file_target().files_count
+
+    @property
+    def is_files_enabled(self) -> bool:
+        return self.get_file_target().is_files_enabled
```
Hercilio1, line 31: Do we want to create this? The ideal plugin architecture would be to use the shared services directly where needed instead of creating an uncontrollable inheritance chain.

baseapp/files/models.py
```diff
@@ -0,0 +1,444 @@
...
+    if not hasattr(sender._meta, "triggers"):
+        sender._meta.triggers = []
+
+    existing = [t.name for t in sender._meta.triggers]
+    if "update_file_target_on_insert" not in existing:
+        sender._meta.triggers.append(file_target_insert_trigger())
+    if "update_file_target_on_update" not in existing:
+        sender._meta.triggers.append(file_target_update_trigger())
+    if "update_file_target_on_delete" not in existing:
+        sender._meta.triggers.append(file_target_delete_trigger())
```
Hercilio1: Instead of doing like that, you can use our pgtrigger_register_default_track from baseapp_core for registering the triggers resolving the Swapped Model (take a look on how it's used in the baseapp_chats).

baseapp/files/models.py
```diff
@@ -0,0 +1,444 @@
...
+    is_files_enabled = models.BooleanField(
+        default=True,
+        verbose_name=_("is files enabled"),
+        help_text=_("Whether files can be attached to the target"),
+    )
+
+    class Meta:
+        abstract = True
+        verbose_name = _("file target")
+        verbose_name_plural = _("file targets")
```
Hercilio1, line 218: Lets add the swappable frag in here. And in the AbstractFile Meta;

Hercilio1 (changes requested):

baseapp/files/models.py
```diff
@@ -0,0 +1,444 @@
...
+class AbstractFile(*file_inheritances, DocumentIdMixin, RelayModel, TimeStampedModel):
+    parent = models.ForeignKey(
+        DocumentId,
+        null=True,
+        blank=True,
+        on_delete=models.CASCADE,
+        related_name="files",
+        verbose_name=_("parent"),
+        help_text=_("The parent document this file belongs to"),
+    )
```
Hercilio1, line 282: On a social media app usually we don't need to worry about alt text, but in a CMS we would need to (for accessability). Do you think we should worry about it? And if so, retrieving a string in the graphql won't be enough

Hercilio1 (approved):

PR #453, opened 2026-08-25
Fix pip-audit: refresh locked dependencies

nossila:
## Problem

`pip-audit` fails on master with **128 known vulnerabilities across 18 packages**. Nothing in the tree changed to cause it — the last passing run on master was 2026-08-07, and the advisory data moved since. Any branch cut from master inherits the red build.

## Fix

`uv lock --upgrade`. Every finding clears from the lock refresh alone — no version floor in `pyproject.toml` needed raising, so this PR only touches `uv.lock` plus one workflow line.

Notable moves:

| Package | From | To |
|---|---|---|
| django | 5.2.14 | 5.2.17 |
| wagtail | 7.3.1 | 7.4.3 |
| pypdf | 6.10.2 | 6.16.2 |
| cryptography | 46.0.7 | 50.0.0 |
| aiohttp | 3.13.4 | 3.14.3 |
| pillow | 12.1.1 | 12.3.0 |
| pyjwt | 2.11.0 | 2.13.0 |
| setuptools | 82.0.1 | 84.0.0 |
| sqlparse | 0.5.5 | 0.6.0 |
| ujson | 5.12.0 | 5.13.0 |

## Also: dropped a stale suppression

The pip-audit step carried `--ignore-vuln CVE-2026-4539`, added for pygments when no fix existed. pygments is now 2.21.0 and the advisory no longer reports against it, so the flag silenced nothing — while still being able to hide that ID indefinitely if it ever came back. Removed. The `Ignored vulnerabilities:` table above the step is kept as an empty scaffold — heading, column headers and separator, with the stale pygments row dropped — so the next suppression is a one-line addition rather than a rebuild of the format.

## Verification

Run locally against the built CI image, not just resolved on paper:

- `pip-audit --skip-editable` — the exact command this workflow runs, **without** any ignore flag → `No known vulnerabilities found`, exit 0
- flake8, isort, black, `ast-grep test`, `ast-grep scan` → all pass
- `uv lock --check` → lock matches `pyproject.toml`, so `uv sync --frozen` is happy

The full pytest run is left to CI here, which runs it in parallel across the matrix rather than single-process.

## Note for consumers

`baseapp-backend-template` pins this submodule and its Jenkins pipeline runs `verifyGitSubmodule(..., 'baseapp-backend', 'master')` for PRs targeting master. Its matching dependency PR will stay red on that stage until this one merges.


nossila:
Follow-up on the verification section: the full suite has now also been run locally against the built CI image, single-process rather than the parallel matrix.

```
1385 passed, 3 skipped, 3494 warnings in 2288.05s (0:38:08)
Required test coverage of 75.0% reached. Total coverage: 82.35%
```

So the lock refresh is green on tests and coverage as well, not just on `pip-audit`. Nothing in the upgrade needed a test or source change — this PR is still `uv.lock` plus the one workflow line.

